PR #1, opened 2022-12-12
Numerous fixes to support CI across platforms.

jsirois:
The main discovery trying to get this running on macOS was that the
PythonBuildStandalone Python distributions can not be counted on for
building platform specific distributions*. This means macOS aarch64
Pants installs tend to fail trying build distributions Pants depends on
like psutil and setproctitle. To avoid this issue, there is a set of
changes to use the binaries.pantsbuild.org pre-built wheels via a
--find-links repo for all installs.

In order to support a fully statically linked `scie-pants` on Linux, the
package crate builds steps are broken up and re-worked as subcommands.
This arrangement allows for building the tools.pex on Ubuntu (glibc),
then the scie-pants scie in an alpine image where it gets fully
statically linked, and then running the integration tests back in
Ubuntu.

* See: https://github.com/indygreg/python-build-standalone/blob/main/docs/quirks.rst#references-to-build-time-paths


jsirois:
So, my plan:
1. This PR to establish CI green baseline.
2. Doc PR with READMEs - 1 for the scie-pants scie (package/scie-pants.lift.json + src/main.rs + tools/src) explaining how this thing works and 1 for the build and test process (package/src/main.rs).
3. Add manual self-update support (`SCIE_BOOT=update scie-pants`).
4. Tag and release 0.1.0

That's all that blocks me saying - try this out for personal use with instructions for how to do that.

For use general in CI I need to address the current use of api.github.com in Pants version installs. This is solved for GitHub users as demonstrated in this PRs CI setup, but that is not a solution for non-GH users. I want to try out the `git2` crate as an alternate way to translate Pants versions to git shas for getting wheels from binaries.pantsbuild.org.

From there it is time to start shipping Pants scies. An extra branch will be added to src/main.rs that, after some threshold version, re-executes a new command added to the manifest and the installer will handle those installs differently: ~just fetching the pants scies.

jsirois:
Final notes:
+ This is setup to auto release just like the scie projects - just need to tag and push tag. The intial releases are {Linux, Mac} x {x86_64, aarch64}; so the scie-pants are ready for Pants Linux aarch64 when it arrives.
+ Please ask away but also note the resources over in a-scie:
  - scie packaging: https://github.com/a-scie/jump/blob/main/docs/packaging.md
  - ptex function: https://github.com/a-scie/ptex/blob/main/README.md

jsirois:
@huonw if you don't mind taking a grok of `tools/src/scie_pants/bootstrap_tools.py`, that's intended to take the place of some of the work you did to support pantsbuild/actions. There will need to be more back and forth there so the scie-pants can communicate relevant cache paths, but that's for later.

@chrisjrn The `tools/src/scie_pants/install_pants.py` takes over the work `one_step_setup.sh` was doing and is a bit more agressive - it edits existing `pants.toml` too using `tomlkit`. Your opinions there would be great. The plan is to follow up with also just plopping down a `pants.toml` on prompt if there is none, but just this for now.

package/scie-pants.lift.json
```diff
@@ -114,14 +114,14 @@
...
           },
           "bootstrap-tools": {
             "description": "Introspection tools for the Pants bootstrap process.",
             "env": {
-              "=PEX_ROOT": "{scie.bindings}/pex_root"
+              "=PEX_ROOT": "{scie.bindings}/pex_root",
+              "=PEX_PYTHON_PATH": "{scie.files.{scie.env.PYTHON}-{scie.platform}}/python/bin/{scie.env.PYTHON}"
             },
             "exe": "{scie.files.{scie.env.PYTHON}-{scie.platform}}/python/bin/{scie.env.PYTHON}",
             "args": [
```
huonw, line 127: Just one small question about `bootstrap_tools.py`, looking for confirmation: on my read of that file in isolation, it _seemed_ to make `bootstrap-cache-key` much harder to use: specifying the `--python-distribution-hash` and `--pants-version` arguments externally seems hard/impossible. However, my read of this section is that users won't have to set them directly because invocations are expected to go via this entry point, and thus have them set automatically.

That is, `PANTS_BOOTSTRAP_TOOLS=1 scie-pants bootstrap-cache-key` (or similar) will be invoking the actual bootstrap tools `main` with the appropriate args, transparent to users?

jsirois: Where the PANTS_VERSION env var is set by this code in the main scie-pants launcher binary just before it re-executes itself to satisfy the request:
https://github.com/pantsbuild/scie-pants/blob/dfb3961cf7122e1ae296e98b80da24164903313c/src/main.rs#L112-L118
https://github.com/pantsbuild/scie-pants/blob/dfb3961cf7122e1ae296e98b80da24164903313c/src/main.rs#L168-L176
https://github.com/pantsbuild/scie-pants/blob/dfb3961cf7122e1ae296e98b80da24164903313c/src/main.rs#L143-L152

jsirois:
One thing I didn't add to this PR's notes was the speed. This is ~33% faster on warm starts (~700ms on my machine -> ~500ms on my machine for current Pants versions) than the existing `./pants` script and supports every one of its features afaict. That script was getting slow and the rust brings us back down to sub 1ms launches of Pants where things slow down again. That next slowdown will be solved by the Pants native client.

sureshjoshi:
> and supports every one of its features afaict

Have you been able to test on in-repo plugins as well?

jsirois:
> Have you been able to test on in-repo plugins as well?

I have not tried, but by the point Pants is running all of this code is out of the way and it's just Pants + IndyGreg's Python. The only known issue is with the latter building c extension; so if the plugin, in repo or 3rdparty, has dependencies with c extensions and no corresponding pre-built wheel - things should go boom.

jsirois:
Another data point that may be useful when evaluating all this, the current largest binary is Linux and that's 8MB fully static. I just experimented adding in the git2 crate fully vendored to support fetching Pants version -> sha and it then grows to 11MB. So that's the ballpark size of the scie-pants. They'll always be just 1 scie-pants binary on a system though since it is responsible for switching between the underlying correct Pants versions per-invocation; so size doesn't matter as much for this scie as it will for the forthcoming Pants scies - of which there will be up to many on a given machine.

sureshjoshi:
Possibly tangential to this, but maybe planned already? I don't think I've seen it discussed explicitly yet. 

Would this lead to the option of having a single, global "pants" executable - and then that exe reads the `pants.toml` (and whatever else) for grabbing versioned Pantses?

As in, would "this" (the pants-scie concept) be a drop-in replacement in each repo for the `./pants` script? Or would this optionally eliminate the `./pants` script altogether?

jsirois:
> Would this lead to the option of having a single, global "pants" executable

That's what this is present tense.

jsirois:
I won't be reccomending use until a few more PRs and the scie-pants scie 0.1.0 being released on GitHub releases, but feel free to try it out:
```
cargo run -p package -- scie
```

That'll plop out `dist/scie-pants-<os>-<arch>` and you can then just cp it to `~/bin/pants` or wherever you choose / with whatever name you choose.

jsirois:
Now a bit more self-describing:
```
$ cargo run -p package -- --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/package --help`
Packages the scie-pants binary.

Usage: package [OPTIONS] <COMMAND>

Commands:
  tools  Builds the `tools.pex` used by the scie-pants scie to perform Pants installs
  scie   Builds the `scie-pants` scie
  test   Builds the `scie-pants` scie and runs it through a series of integration tests
  help   Print this message or the help of the given subcommand(s)

Options:
      --target <TARGET>        Override the default --target for this platform.
      --ptex <PTEX>            Instead of using the released v0.6.0 ptex, package ptex from the ptex project repo at this directory.
      --scie-jump <SCIE_JUMP>  Instead of using the released v0.6.0 scie-jump, package the scie-jump from the scie-jump project repo at this directory.
      --update-lock            Refresh the tools lock before building the tools.pex
      --dest-dir <DEST_DIR>    The destination directory for the chosen binary and its checksum file. [default: dist]
  -h, --help                   Print help information
  -V, --version                Print version information

Subcommand help:
```
$ cargo run -p package -- scie --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/package scie --help`
Builds the `scie-pants` scie

Usage: package scie [OPTIONS]

Options:
      --tools-pex <TOOLS_PEX>  The location of the pre-built tools.pex to use. By default, the tools.pex is built fresh.
  -h, --help                   Print help information
```

File: package/scie-pants.lift.json (hunk 1 above)
Comment (huonw) [reply package/scie-pants.lift.json:127]: Very cool. Thanks for confirming!

File: tools/src/scie_pants/install_pants.py [hunk 2]
```diff
@@ -19,54 +22,156 @@
...
+
+
+def determine_tag_version(
+    ptex: Ptex, pants_version: str, find_links_dir: Path, github_api_bearer_token: str | None = None
+) -> ResolveInfo:
+    github_api_url = (
+        "https://api.github.com/repos/pantsbuild/pants/git/refs/tags/"
+        f"release_{urllib.parse.quote(pants_version)}"
+    )
+    headers = (
```
jsirois, line 123: Although this works, it's a bit titchy. It's OK if:
+ You have a ~/.netrc with an api.github.com entry (my personal case).
+ You export PANTS_BOOTSTRAP_GITHUB_API_BEARER_TOKEN (the case for this green CI)
+ You are a normal user (IP address) whose 60 unauthenticated request per hour quota is plenty (2 API calls per Pants version that needs to be installed).

It's not OK if your Pants using repo is hosted in, say, GitLab and you want to set up CI and you never go near GitHub. We shouldn't be asking you to create a GitHub account and generating a token to use just to use Pants - the two (should)have nothing to do with each other.

Mitigating all this is that this is just a bridge until Pants scie's start shipping (scie-pants does not install pants-scies right yet as you should notice).

It's easy enough for me to snapshot all tag -> sha in a json file say and embed that here. That though goes stale the minute we release 2.16.0.dev3, etc; so api requests are needed for those. I think the best path forward may be to backfill and start storing a https://binaries.pantsbuild.org/wheels/pantsbuild.pants/<tag> file in S3 as part of the release process for stable (tagged) releases. That would be a file that contains just the release tip commit sha. I could backfill those files for all existing tags and add that step to the currently active Pants release branches (main with cherry-picks back though 2.14.x).

chrisjrn: I agree on the release index (noting that using the GitHub API for supporting `PANTS_SHA` is probably fine long-term)

package/src/main.rs
```diff
@@ -27,26 +31,116 @@ const CARGO_MANIFEST_DIR: &str = env!("CARGO_MANIFEST_DIR");
...
 lazy_static! {
-    static ref OS_ARCH: String = format!(
-        "{os}-{arch}",
-        os = env::consts::OS,
-        arch = env::consts::ARCH
-    );
+    static ref CURRENT_PLATFORM: Platform = Platform::current().unwrap();
+}
+
+macro_rules! log {
```
stuhood, line 94: Since dependency count doesn't seem to be an issue for this binary, I'm surprised that this isn't using an existing logging package (I could understand wanting to keep the final product thinner though).

stuhood (approved):
Thanks John!

package/src/main.rs
jsirois: Well, these are all INFO and I differentiate build steps from ITs with Cyan and Magenta. Certainly dumb, but I don't know how to do this with the logging package and this was easier to figure out. The driver behind this apparent ~frivolity was me myself not understanding what was going on. It's a complex build, and RUST_LOG=debug|trace is useful but ~insane and only if you're in the weeds.

jsirois: I did find simpler color printing crates, but they had ~1k uses. I've been trying to stick to Tolnays, BurntSushis, Chrichton's, etc.

tools/src/scie_pants/install_pants.py
jsirois: Yeah, I think the index / tag files in S3 is the way to go. For the `PANTS_SHA` there is no `api.github.com` use and so no rate limit issue at all. We just scrape the raw VERSION file from `github.com` in that case.

jsirois: https://github.com/pantsbuild/scie-pants/blob/49399e7cc2c5fcd5f04d4a5441e64d063844bf21/tools/src/scie_pants/install_pants.py#L133-L140

jsirois: All right. The tag database is in now at ~43KB and ~30ms to load and lookup. That plus the work in https://github.com/pantsbuild/pants/pull/17801 and a one-time S3 back-fill will eliminate all GitHub API use.

jsirois:
Thanks for your reviews. I welcome more feedback and reports of tests using the `scie-pants` yourself to work in Pants repos, but I'll also plow ahead here tomorrow before noon Pacific to march towards a release of 0.1.0 by 12/14 hopefully.

PR #13, opened 2022-12-16
Add self-update capability to scie-pants.

### LINKED ISSUE #8: Support scie-pants self-updating.
There should be basic support for getting the latest stable release of scie-pants from the GitHub releases. This should be as simple as `SCIE_BOOT=update scie-pants` or similar for now. Eventually it would be good to add in every N timeunit checking for new releases etc, but just `SCIE_BOOT=update scie-pants` is a step up.

jsirois:
You can now execute `SCIE_BOOT=update scie-pants` to upgrade if possible
or use `SCIE_BOOT=update scie-pants <version>` to switch to the given
scie-pants version if it is available.

Fixes #8

### DIFF
--- a/tools/src/scie_pants/update_scie_pants.py
+++ b/tools/src/scie_pants/update_scie_pants.py
@@ -69,14 +69,26 @@ class Release:
         return None
 
 
+def fetch_github_api(ptex: Ptex, path: str, github_api_bearer_token: str | None = None) -> Any:
+    headers = (
+        {"Authorization": f"Bearer {github_api_bearer_token}"} if github_api_bearer_token else {}
+    )
+    return ptex.fetch_json(f"{GITHUB_API_BASE_URL}/{path}", **headers)
+
+
 class ReleaseNotFoundError(Exception):
     pass
 
 
-def get_release(ptex: Ptex, platform: str, version: str) -> Release:
+def get_release(
+    ptex: Ptex, platform: str, version: str, github_api_bearer_token: str | None = None
+) -> Release:
     try:
         release_data = cast(
-            dict[str, Any], ptex.fetch_json(f"{GITHUB_API_BASE_URL}/releases/tags/v{version}")
+            dict[str, Any],
+            fetch_github_api(
+                ptex, f"releases/tags/v{version}", github_api_bearer_token=github_api_bearer_token
+            ),
         )
     except (CalledProcessError, OSError) as e:
         raise ReleaseNotFoundError(str(e))
@@ -88,8 +100,13 @@ def get_release(ptex: Ptex, platform: str, version: str) -> Release:
     return release
 
 
-def find_latest_production_release(ptex: Ptex, platform: str) -> Release | None:
-    releases = cast(list[dict[str, Any]], ptex.fetch_json(f"{GITHUB_API_BASE_URL}/releases"))
+def find_latest_production_release(
+    ptex: Ptex, platform: str, github_api_bearer_token: str | None = None
+) -> Release | None:
+    releases = cast(
+        list[dict[str, Any]],
+        fetch_github_api(ptex, "releases", github_api_bearer_token=github_api_bearer_token),
+    )
 
     latest_releases = []
     for release_data in releases:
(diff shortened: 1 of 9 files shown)


jsirois:
Lots of scie-packaging concepts are exercised here with a new boot command and a a scie binding command to support it. If any of this looks like greek - particularly grokking the flow, it's a good time to speak up. The hope is someone else starts adding features / fixing bugs here soon. Contribution docs amongst others coming later today...

kaos:
I ought to take a deep plunge into the whole scie-jump scie-pants world to get proper bearings here.. :)

sureshjoshi:
I'm going to be going through a large backlog of catching up on this during the holidays.

December turned out to be quite the month of trying to get client work done before everyone checks out for the holidays.

jsirois:
Ah, self-update test 403. I need to wire up the GHA token: https://github.com/pantsbuild/scie-pants/actions/runs/3715927556/jobs/6301665892#step:7:214

Coming up...

jsirois:
I'm going to shuffle this one along and try to get a release out this evening, and then a second release such that this can be vetted for real. If I submit before your review, know that I'm still happy to get your feedback and will follow up.

tools/src/scie_pants/update_scie_pants.py
```diff
@@ -0,0 +1,288 @@
...
+    # Mark the binary as executable. This is needed on Unix but not on Windows, where its harmless.
+    binary.chmod(0o755)
+
+    # We first move the running scie before replacing it. This satisfies Windows constraints
+    # surrounding manipulation of running binaries.
+    backup = scie.rename(scie.with_suffix(".bak"))
+    try:
+        shutil.move(str(binary), str(scie))
+    except OSError:
+        warn(f"A backup is saved in {backup}")
```
stuhood, line 177: Why not try moving the backup back here?

tools/src/scie_pants/update_scie_pants.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+def main() -> NoReturn:
+    parser = ArgumentParser()
+    get_ptex = Ptex.add_options(parser)
+    parser.add_argument(
+        "--platform",
+        required=True,
+        # The current platform tag (<OS>-<ARCH>)
+        help=argparse.SUPPRESS,
```
stuhood, line 202: Why are these suppressed?

stuhood (approved):
Thanks!

tools/src/scie_pants/update_scie_pants.py
jsirois: They are curried by the lift manifest - user's do not pass them, so if they showed up in help that would be confusing to users.

tools/src/scie_pants/update_scie_pants.py
jsirois: I think once you start to run into issues its best to hands off and give clear info. If a _second_ error happened trying to do too much ... that's just a mess. Right now you get a clear message as to where the binary was moved to. It's very easy to just cut/paste move it back.

jsirois: This will definitely require some playing around with. To even arrive at the move first then replace required a bunch of messing around testing on Windows, which is what this is all dealing with. Simple on Unix.

tools/src/scie_pants/update_scie_pants.py
jsirois: It hides all this garbage which the user has no change of specifying:
https://github.com/pantsbuild/scie-pants/blob/91aa50b517523fed0d054831274351e022696029/package/scie-pants.lift.json#L147-L158

Instead, they see only the unbound help:
```
$ SCIE_BOOT=update dist/scie-pants-linux-x86_64 --help
usage: tools.pex [-h] [version]

positional arguments:
  version     The version of scie-pants to update to; defaults to the latest stable version

optional arguments:
  -h, --help  show this help message and exit
```

jsirois: I need to fix that "tools.pex" prog name leak, that also means nothing to an end user, but that's for follow up.

PR #21, opened 2022-12-17
Fix up self-update to skip current version.

jsirois:
Previously the scie-pants would update itself to its same version instead of short circuiting.


benjyw (approved):

jsirois:
This is enough to finish some docs and announce for testing. The self-update appears to be robust and you can specify a version to downgrade instead which works as well.

Issue #7, opened 2022-12-14
Document the scie-pants repo


jsirois:
This should include:
+ [x] General Readme: directed at end users for the most part.
+ [ ] Design & Flow: directed at developers of scie-pants.
+ [ ] Build System: directed at developers of scie-pants / folks wanting to modify the package helper crate.
+ [x] Contributing: directed at new developers.
+ [x] Release: directed at developers performing releases.
+ [x] Changes: The user-facing high-level changelog written by developers and used by the automated release process for the Releases page and the Announcements discussions.

PR #23, opened 2022-12-18
The outer layer of dev docs.

jsirois:
Additional more detailed docs on the basic structure of scie-pants and
its main process execution flows is coming separately.

Work towards #7

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,81 @@
+# Contributing
+
+The `scie-pants` binary is intended to be the next generation Pants launcher.
+
+Like the current `./pants` bash  script launcher, it determines the required Pants version for the
+current Pants-using project and launches it, installing that Pants version if needed.
+
+Unlike the current `./pants` bash script, the `scie-pants` binary:
++ Provides a guaranteed CPython 3.8+ interpreter appropriate to the Pants version being installed.
++ Is intended for use globally as a binary installed on the `PATH`.
+
+When invoked, the `scie-pants` binary searches from the `CWD` of its invocation up through ancestor
+directories to find the `pants.toml` signalling a Pants-using project. It then parses `pnats.toml`
+to discover the Pants version required and launches that Pants version. Since the `scie-pants`
+launcher is a native binary, this process is very fast, typically ~one millisecond, in the hot path
+where the needed Pants version is already installed.
+
+The slow path, where a new Pants version needs to be installed or Pants bootstrap utility functions
+are invoked, is not performance sensitive and can benefit from the Python 3.8+ interpreter the
+`scie-pants` launcher guarantees. As such, the Pants installer, bootstrap tools and more are written
+in Python 3 and embedded in the `scie-pants` scie as a single `tools.pex` [conscript busy-box](
+https://pypi.org/project/conscript/) with multiple console script entry points. This allows for the
+core logic of the slow path to be complex but manageable with type-checking, unit testing and all
+the power Python generally provides.
+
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,81 @@
...
+
+Like the current `./pants` bash  script launcher, it determines the required Pants version for the
+current Pants-using project and launches it, installing that Pants version if needed.
+
+Unlike the current `./pants` bash script, the `scie-pants` binary:
++ Provides a guaranteed CPython 3.8+ interpreter appropriate to the Pants version being installed.
++ Is intended for use globally as a binary installed on the `PATH`.
+
+When invoked, the `scie-pants` binary searches from the `CWD` of its invocation up through ancestor
+directories to find the `pants.toml` signalling a Pants-using project. It then parses `pnats.toml`
```
kaos, line 13: Typo, `pnats.toml`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,81 @@
...
+code when iterating on it.
+
+When you're ready to get additional eyes on your changes, submit a [pull request](
+https://github.com/pantsbuild/scie-pants/pulls).
+
+## Guiding Principles
+
+There are just a few guiding principles to keep in mind as alluded to above:
++ The `scie-pants` hot path should be fast: It currently launches Pants in ~one millisecond and that
+  should hold steady.
```
kaos, line 76: Could we have/are there any perf tests that may run automatically/in CI to ensure this will hold true over time?

kaos (approved):

CONTRIBUTING.md
jsirois: We should. I'm not sure exactly how to do it yet. Pants gets in the way since its slow. There are 3 separate binary executions in the hot path:
1. The scie-jump launches the embedded scie-pants.bin binary as the default command.
2. The scie-pants binary finds pants.toml, parses the pants_version and determines the proper Python version and then re-launches itself specifying a few env vars to convey what it found and SCIE_BOOT=pants to launch the pants command contained in the scie.
3. The scie-jump launches the pants command.
4. The pants console script in the previously installed Pants venv runs.

Each of the 1st 3 steps takes ~300us. The 4th step takes 500ms min (for pants -V for example).

Perhaps there is a way to shim out 4 with something other than the pants console script that runs in consistent time on the order of 1ms or less (like `bash -V`). Perhaps even that could be avoided.

Right now I just run with RUST_LOG=debug which gives the timings I just described for the 1st 3 steps in log lines.

CONTRIBUTING.md
jsirois: Fixed.

CONTRIBUTING.md
jsirois: #24 now tracks this.

jsirois: I will say though to not really worry about this if it feels scary when considering contribution. This is both because the actual scie-pants binary does almost nothing; so its hard to make it slow, and because most feature work will just be, at most, plumbing reading of new toml values from an already priced in pants.toml parse and setting new env vars for sub-processes - both of which are really hard to make slow in non-obvious ways.

jsirois: The slowest thing it does actually, by almost an order of magnitude, is handle `.pants.bootstrap`! So this is step 2 which takes ~300us with no `.pants.bootstrap` and adding in a ~minimal GIT_COMMIT `./pants.bootstrap` adds >2000us.

PR #27, opened 2022-12-19
Achieve parity with one_step_setup and then some.

jsirois:
Pants will now configure `pants_version` to be latest stable if
`pants.toml` is present but `pants_version` is not. Further, it will
establish a `pants.toml` if none is present. If the CWD is nested in
a `.git` dir, the `pants.toml` is created there instead.

To support configuration in these cases, break out a `configure`
binding to handle prompting and version selection separate from the
`install` binding.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,7 +8,7 @@ use anyhow::{anyhow, bail, Context, Result};
 use build_root::BuildRoot;
 use log::{info, trace};
 use logging_timer::{time, timer, Level};
-use pyver::PackageVersion;
+use uuid::Uuid;
 
 use crate::config::PantsConfig;
 use crate::pants_bootstrap::PantsBootstrap;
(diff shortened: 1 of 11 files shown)


src/main.rs
```diff
@@ -73,68 +73,60 @@ impl Process {
...
     } else if env_pants_sha.is_none() {
-        pants_config.package_version()
+        configured_pants_version
     } else {
         None
     };
 
-    let debugpy_version = pants_config.debugpy_version().unwrap_or_default();
-
-    let python = if let Some(version) = pants_version {
```
jsirois, line 122: This logic moved into the `configure` binding. Any move like this is good - the less the scie-pants binary does, the better. It's in the hot path and ideally should know as little about Pants as possible.

tools/src/scie_pants/configure_pants.py
```diff
@@ -0,0 +1,136 @@
...
+    cwd = os.getcwd()
+    buildroot = Path(cwd)
+    if shutil.which("git"):
+        result = subprocess.run(
+            args=["git", "rev-parse", "--show-toplevel"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.DEVNULL,
+        )
+        if result.returncode == 0:
+            buildroot = Path(os.fsdecode(result.stdout.strip()))
```
jsirois, line 51: This could conceivably be annoying for some strange case where you want pants.toml, for some reason, set up under the ambient `.git`. It seems to me that situation is almost by definition involving a user who doesn't need one step setup! That said, it may be good to follow up here and provide an extra level of prompting that allows build root selection override. I was hoping to keep this simple for initial release / announce tomorrow though and hopefully let others with opinions about UX chime in by doing - this stuff needs more contributors.

jsirois: The UX for all this can be read from the new IT logs:
https://github.com/pantsbuild/scie-pants/actions/runs/3728396083/jobs/6323372959#step:6:184

Probably easier to grok just trying for yourself. I just copy `dist/scie-pants-XYZ` to `~/bin/pants` and go play around executing `pants`.

benjyw: That `pants.toml` under `.git` case sounds atrocious, and I am fine making it hard to do for now.

src/main.rs
```diff
@@ -154,19 +146,34 @@ fn get_pants_process() -> Result<Process> {
...
+        ));
+    }
     if let Some(version) = pants_version {
-        env.push(("PANTS_VERSION".into(), version.original.clone().into()));
+        env.push(("PANTS_VERSION".into(), version.into()));
+    } else {
+        // Ensure the install binding always re-runs when no Pants version is found so that the
+        // the user can be prompted with configuration options.
+        env.push((
+            "PANTS_VERSION_PROMPT_SALT".into(),
```
benjyw, line 174: Neat trick

benjyw (approved):

src/main.rs
jsirois: Something something about all systems eventually becoming workflow systems ...

PR #43, opened 2022-12-20
Support fetching PBS CPython behind firewalls.

jsirois:


jsirois:
Of course this does not help with installing Pants itself. Legacy Pants is still spread out everywhere on PyPI and in our S3 bucket. Just as today, folks can use PIP_STEERING_ENV_VARS or else their `pip.conf`, if present, will be picked up by Pip.

benjyw (approved):

PR #47, opened 2022-12-22
Cache ptex and scie-jump downloads.

jsirois:


sureshjoshi:
What's the use case where this cache is used? After a successful download, but an interrupted extraction?

sureshjoshi:
I ask because I thought the extracted digest effectively acts as an unpacked result that future ptex/jump calls refer to?

jsirois:
@sureshjoshi exactly right. These components are used to build a scie and need to be as-advertised. I can only blame this on trying code during layovers 🤦

sureshjoshi:
Okay, and it looks like it's not co-located with the `nce`, correct?

`macOS.../Library/Caches/scie-pants/dev/downloads/...`?

jsirois:
Yeah, this is not a scie app, it's the build process for one using it's own bespoke cache of downloaded ptex and scie-jump binaries to build the scie with.

jsirois:
Basically I need to get around to working on a-scie/lift, that blank repo is intended to be a high level tool (shipped as a scie, written in Python) that is what normal people use to build scie apps. It would handle getting scie-jump and ptex for you to build the scie with and lots of code here would go away.

kaos:
so many word-puns 😂

sureshjoshi:
👍🏽  Just confirming my current understanding

sureshjoshi (approved):

jsirois:
@sureshjoshi PTAL at the last commit - things were buggy in the way you suggest before that!

package/src/main.rs
```diff
@@ -444,33 +444,27 @@ fn fetch_a_scie_project(
...
-    if 0 == std::fs::metadata(&cached_binary)
-        .map_err(|e| {
-            Code::FAILURE.with_message(format!(
-                "Failed to retrieve metadata for {path}: {e}",
-                path = cached_binary.display()
-            ))
-        })?
-        .len()
-    {
+    if !target_dir.exists() {
```
sureshjoshi, line 462: What's was the atomicity failure vs the previous commits? I thought based on the previous commits that the `target_dir` is created before this?

jsirois:
Noting after this that the only thing slow about the dev cycle is the `tools.pex` build step. That we can't use Pants for (bootstrapping circularity here! There is no Pants launcher yet, we're building it!); so I'll need to look into some bespoke caching for that step as a follow up.

package/src/main.rs
sureshjoshi: >  the partial binary would be there

Ahhh, okay, I thought this process was also atomic. That makes sense now.

sureshjoshi (approved):

package/src/main.rs
sureshjoshi: My original thought was not about corruption, but rather just network caching - I didn't think we could half unpack a digest

jsirois: Aha, ok.

Issue #35, opened 2022-12-19
Create a `get-pants` script and site.


jsirois:
We'll probably need a `.sh` version and a `.ps1` version that detects the current OS-ARCH and downloads the latest stable scie-pants, adds to $PATH, etc. A gold standard is probably https://rustup.rs/. Critically, it would be good to be able to have the script checked in to Pants repos. This would allow new users of a repo using Pants to run `./get-pants.sh` to get up and running by installing `scie-pants` (as `pants`) on the PATH.

benjyw:
How do we feel about `pantsup` as a name, both in analogy to rustup and for the pun?

jsirois:
This will get checked in to a repo and the actual update command is `SCIE_BOOT=update pants` as documented here:
https://github.com/pantsbuild/scie-pants/blob/2f2b2365c31b45d2fad792810df4855f65b088d2/get-pants.sh#L147-L149

So - I don't really care, but those are your logical / reality constraints on word use.

benjyw:
Hmm, I hadn't thought about checking it in to the repo. That seems weird to me, but maybe? Checking in `./pants` is different in my mind, as you must have that in every repo to do anything. 

E.g., I wouldn't check `rustup` in to the repo.

kaos:
But we could have a recommendation for a one-liner to check-in:
```shell
# get-pants.sh --downloads pantsup.sh and runs it
bash -c "$(curl ...)"
```

jsirois:
@kaos why check in a 1 liner instead of the script? The script is basically a robust 1 liner.

jsirois:
To be super clear (I cannot find the Slack user convo):

I personally would not check in the script. A user, however, pointed out they explicitly liked that there was a checked in script since it made it easy to on-board new devs to the repo with it checked in. They felt "run this script" was easier than separate instructions for how to install the binary 1 time. Taking that concern / perceived benefit seriously, they'd still want this presumably.

kaos:
> @kaos why check in a 1 liner instead of the script? The script is basically a robust 1 liner.

Oh, I'm in no way against checking in the entire script. It was merely a reaction as to an alternative option to Benjy's remark:
> E.g., I wouldn't check rustup in to the repo.

[...]

PR #48, opened 2022-12-23
Add get-pants.sh script.

jsirois:
This will be made available on a rustup.rs-like site and can be
checked-in to a repo to allow new users to on-board easily. The first
cut does not include a Power Shell variant, although it does work with
Git bash for Windows.

Work towards #35

### DIFF
--- a/get-pants.sh
+++ b/get-pants.sh
@@ -110,7 +110,8 @@ function install_from_url() {
   )
   rm "${workdir}/"*.sha256
   if [[ "${OS}" == "macos" ]]; then
-    install -d -m 755 "${workdir}/"* "${dest}"
+    mkdir -p "$(dirname "${dest}")"
+    install -m 755 "${workdir}/"* "${dest}"
   else
     install -D -m 755 "${workdir}/"* "${dest}"
   fi
--- a/get-pants.sh
+++ b/get-pants.sh
@@ -1,0 +1,211 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+function log() {
+  echo -e "$@" >&2
+}
+
+function warn() {
+  log "WARNING: " "$@"
+}
+
+function die() {
+  log "$@"
+  exit 1
+}
+
+function check_cmd() {
+  local cmd="$1"
+  command -v "$cmd" >/dev/null || die "This script requires the ${cmd} binary to be on the PATH."
+}
+
+_GC=()
+
+function gc() {
+  if [[ -z "${NO_GC:-}" ]]; then
+    if (($# > 0)); then
+      check_cmd rm
+      _GC+=("$@")
+    else
+      rm -rf "${_GC[@]}"
+    fi
+  fi
+}
+
+trap gc EXIT
+
+check_cmd uname
+
+function calculate_os() {
+  local os
+
+  os="$(uname -s)"
+  if [[ "${os}" =~ [Ll]inux ]]; then
+    echo linux
+  elif [[ "${os}" =~ [Dd]arwin ]]; then
+    echo macos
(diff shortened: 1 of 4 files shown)


jsirois:
For example, in setup mode - no Pants, no Pants repo:
```
$ mkdir /tmp/foo
$ cd /tmp/foo
$ /usr/bin/env bash -c "$(curl -fsSL file:///home/jsirois/dev/pantsbuild/jsirois-scie-pants/get-pants.sh)"
Downloading and installing scie-pants release ...
Installed scie-pants 0.2.0 from https://github.com/pantsbuild/scie-pants/releases/latest/download/scie-pants-linux-x86_64 to /home/jsirois/bin/pants
Launching pants ...
No Pants configuration was found at or above /tmp/foo.
Would you like to configure /tmp/foo as a Pants project? (Y/n): y
Fetching latest stable Pants version since none is configured
Creating /tmp/foo/pants.toml and configuring it to use Pants 2.14.0
10:09:13.31 [INFO] Initializing scheduler...
10:09:13.39 [INFO] Scheduler initialized.
10:09:13.40 [WARN] Please either set `enabled = true` in the [anonymous-telemetry] section of pants.toml to enable sending anonymous stats to the Pants project to aid development, or set `enabled = false` to disable it. No telemetry sent for this run. An explicit setting will get rid of this message. See https://www.pantsbuild.org/v2.14/docs/anonymous-telemetry for details.
2.14.0
$ cat pants.toml
[GLOBAL]
pants_version = "2.14.0"
```

get-pants.sh
```diff
@@ -0,0 +1,212 @@
...
+    if (($# > 0)); then
+      check_cmd rm
+      _GC+=("$@")
+    else
+      rm -rf "${_GC[@]}"
+    fi
+  fi
+}
+
+trap gc EXIT
```
kaos, line 36: neat trick 👀

get-pants.sh
```diff
@@ -0,0 +1,212 @@
...
+  if [[ "${os}" =~ [Ll]inux ]]; then
+    echo linux
+  elif [[ "${os}" =~ [Dd]arwin ]]; then
+    echo macos
+  elif [[ "${os}" =~ [Ww]in|[Mm][Ii][Nn][Gg] ]]; then
+    # Powershell reports something like: Windows_NT
+    # Git bash reports something like: MINGW64_NT-10.0-22621
+    echo windows
+  else
+    die "Integration tests are not supported for this operating system (${os})."
```
kaos, line 53: This message gives me pause.. if this script is run by a user on an unsupported OS, and they see "Integration tests are not supported for this operating system ...", well if they want to get up and running with Pants, what has integration tests got to do with it? (I see that may be the reason why it's not available to them, but I find it a not very helpful message to convey)

kaos (approved):

get-pants.sh
jsirois: Copy-pasta: https://github.com/a-scie/jump/blob/main/examples/run.sh
Thanks for catching that.

get-pants.sh
```diff
@@ -0,0 +1,212 @@
...
+function calculate_arch() {
+  local arch
+
+  arch="$(uname -m)"
+  if [[ "${arch}" =~ x86[_-]64 ]]; then
+    echo x86_64
+  elif [[ "${arch}" =~ arm64|aarch64 ]]; then
+    echo aarch64
+  else
+    die "Pants is not supported for this chip architecture (${arch})."
```
jsirois, line 129: This one I had fixed, just missed the OS message.

kaos: ah, yes noticed this one after writing the other comment... 👍🏽

Issue #57, opened 2023-01-13
Would this be a good candidate for a Homebrew formula?


sureshjoshi:
I install most of my Mac tooling via Homebrew, just wondering if someone would know better if this is a candidate for that?

jsirois:
I don't know the answer, but I do hope it's no! There is definitely a tension here. Supporting homebrew - in my mind - means supporting  apt / dpkg, yum / dnf / rpm and ... msi? Nice for users of these package formats / package managers, but it still leaves people in the cold (pacman) and is a lot of new knowledge / work / maintenance on this side of the fence.

sureshjoshi:
Ahh, fair point - I selfishly was only thinking about one of my workflows :)

jsirois:
I won't object to a contribution of course. I personally won't devote time to learning 4 package formats and building this all out, but I will gladly review folks that want to contribute one or more.

sureshjoshi:
Fair enough, that could be community run I suppose.

kaos:
+1 for community supported packaging, and I'd be happy to contribute to the homebrew tap. (been having the same idea floating about but not yet surfacing..) :)

jsirois:
That would be excellent then. Maybe you two could take that over.

kaos:
Hmm... I'm not sure if this will be accepted into the core homebrew tap, due to this issue for new formulas:
```
* GitHub repository not notable enough (<30 forks, <30 watchers and <75 stars)
```

It is easy enough to setup a custom tap, just slightly annoying when installing as it'll be more difficult to discover.

kaos:
https://github.com/Homebrew/homebrew-core/pull/120557

kaos:
As an example what that may look like, here's an small example tap: https://github.com/kaos/homebrew-shell

jsirois:
Is there no support for pre built binaries? It's pretty atrocious to go through the completely un-necessary work of re-building a pre-built binary. The hash will even be the same ... it's literally burning the planet down for absolutely no purpose.

kaos:
> Is there no support for pre built binaries? It's pretty atrocious to go through the completely un-necessary work of re-building a pre-built binary. The hash will even be the same ... it's literally burning the planet down for absolutely no purpose.

Ah, there is. That's the `cask` thing. I've always associated casks with "Apps", that is, mostly GUI applications, but what it really means is "install binary distribution".

I've created a cask for pants here: https://github.com/pantsbuild/homebrew-pants

[...]

PR #58, opened 2023-01-14
Don't invoke rustup add unless needed.

jsirois:


jsirois:
This is in response to @kaos's work in #57 to get a Homebrew package set up.

package/src/main.rs
```diff
@@ -403,6 +405,56 @@ impl BuildContext {
...
+                target = self.target
+            );
+            execute(Command::new("rustup").args(["target", "add", &self.target]))?;
+            self.target_prepared = true;
+        }
+        Ok(())
+    }
+
+    fn obtain_ptex(&mut self, dest_dir: &Path) -> Result<PathBuf, Exit> {
+        if self.ptex_repo.is_some() {
```
jsirois, line 422: I was not clever enough to be able use an `if let Some(...) = ...` binding here and satisfy the borrow checker; so this safe, but unsatisfying is_some / unwrap unconditionally a few lines lower combo.

jsirois: Got smarter, Cell allows for a partially mutable struct. Fixed.

kaos (approved):
Thanks 👍🏽

PR #65, opened 2023-01-24
Fix `.pants.bootstrap` handling to only gather vars.

### LINKED ISSUE #63: Exported shell functions create noise when running `pants`
I define a utility functions in my `~/.bash_profile`, and `export -f` it:
```shell
function run_testdb () {
    docker run --rm -it \
        -e POSTGRES_HOST_AUTH_METHOD=trust \
        -e POSTGRES_DB=django \
        -e POSTGRES_USER=django_user \
        -p 5432:5432 \
        public.ecr.aws/docker/library/postgres:12 \
        -c max_locks_per_transaction=512 \
        -c fsync=false \
        -c synchronous_commit=off \
        -c full_page_writes=false \
        -c shared_buffers=256MB \
        -c autovacuum=off
}
export -f run_testdb
```

When I run the `scie-pants` binary in a shell after loading this profile, I see logs:
```
Could not interpret env entry run_testdb () . Skipping.
Could not interpret env entry { . Skipping.
Could not interpret env entry }. Skipping.
Could not interpret env entry run_testdb () . Skipping.
Could not interpret env entry { . Skipping.
Could not interpret env entry }. Skipping.
```

I'm not sure if it's possible to detect & silently skip exported functions (while preserving the log for non-functions). If not, it'd be nice to hide these behind some sort of "debug" logging option.

jsirois:
Previously everything reported by the `set` builtin was parsed and this
could include functions and other exotica leading to un-needed warnings.

Fixes #63

### DIFF
--- a/src/pants_bootstrap.rs
+++ b/src/pants_bootstrap.rs
@@ -30,12 +30,25 @@ impl PantsBootstrap {
                 "-euo",
                 "pipefail",
                 "-c",
-                format!(
-                    r#"set >&2; source "{pants_bootstrap}" >"{capture}" 2>&1; set"#,
-                    pants_bootstrap = pants_bootstrap.display(),
-                    capture = capture.path().display(),
-                )
-                .as_str(),
+                // N.B.: The `set` are used to print the current variables. The set -o posix are
+                // used to toggle restriction of `set` to printing only variable names and not more
+                // exotic items like functions. We ensure our modifications are unset before
+                // sourcing the .pants.bootstrap by using a sub-shell so that that script can use
+                // full bash-isms as has been the contract established in the original `./pants`
+                // script.
+                // N.B.: We take care to ensure the bash script here is a single line since the
+                // script contents are re-capitulated in the BASH_EXECUTION_STRING env var and a
+                // multiline env var will have its trailing lines fail to parse as env vars in most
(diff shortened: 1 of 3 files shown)


src/pants_bootstrap.rs
```diff
@@ -30,11 +30,27 @@ impl PantsBootstrap {
...
+                // exotic items like functions. We ensure our modifications are unset before
+                // sourcing the .pants.bootstrap by using a sub-shell so that that script can use
+                // full bash-isms as has been the contract established in the original `./pants`
+                // script.
+                // N.B.: We take care to ensure the bash script here is a single line since the
+                // script contents are re-capitulated in the BASH_EXECUTION_STRING env var and a
+                // multiline env var will have its trailing lines fail to parse as env vars in most
+                // cases.
+                [
+                    r#"(set -o posix; IFS=$'\0'; set >&2); "#,
```
danxmoran, line 44: I don't fully grok why the `set >&2` is needed as part of this. Does it still apply properly here when run in a sub-shell?

jsirois: The overall process is to diff a before and after set of shell variables. I use STDERR to capture before and STDOUT to capture after.

danxmoran (approved):

PR #75, opened 2023-01-26
Delegate to `./pants` when `[DEFAULT] delegate_bootstrap=true`.

### LINKED ISSUE #33: Special-case the Pants repo.
Until the `scie-pants` binary can actually be used by Pants developers in the Pants repo, it would be good if it could detect it was running in the Pants repo.

Perhaps with:
```toml
[bootstrap]
delegate = true
```

And then just run the `./pants` script in the Pants repo.

kaos:
The choice of config section for the `pants.toml` config file was made to have something that won't choke the regular Pants configuration verification and also not show up in any online help.

Closes #33

### DIFF
--- a/package/scie-pants.lift.json
+++ b/package/scie-pants.lift.json
@@ -118,6 +118,10 @@
               "{scie.bindings.configure:PANTS_SHA_FIND_LINKS}"
             ]
           },
+          "pants-dev": {
+            "description": "Runs Pants from a local source checkout of the pantsbuild/pants repository, for development.",
+            "exe": "{scie.env.PANTS_BUILDROOT_OVERRIDE}/pants"
+          },
           "bootstrap-tools": {
             "description": "Introspection tools for the Pants bootstrap process.",
             "env": {
--- a/package/scie-pants.lift.json
+++ b/package/scie-pants.lift.json
@@ -118,10 +118,6 @@
               "{scie.bindings.configure:PANTS_SHA_FIND_LINKS}"
             ]
           },
-          "pants-dev": {
-            "description": "Runs Pants from a local source checkout of the pantsbuild/pants repository, for development.",
-            "exe": "{scie.env.PANTS_BUILDROOT_OVERRIDE}/pants"
-          },
           "bootstrap-tools": {
             "description": "Introspection tools for the Pants bootstrap process.",
             "env": {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ${{ env.SCIE_PANTS_DEV_CACHE }}
-          key: ${{ runner.os }}-${{ runner.arch }}-scie-pants-v1
+          key: ${{ runner.os }}-${{ runner.arch }}-scie-pants-v2
       - name: Build, Package & Integration Tests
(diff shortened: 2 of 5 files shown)


kaos:
I'm not sure how/where to write tests for this, but it seems to do what I want when tested on the Pants repo.

I have observed though that sometimes it seems like it's attempting to connect with what I suspect is debugpy, but I've no clue as to why that would be.

package/scie-pants.lift.json
```diff
@@ -118,6 +118,10 @@
               "{scie.bindings.configure:PANTS_SHA_FIND_LINKS}"
             ]
           },
+          "pants-dev": {
```
jsirois, line 121: So a lift entry means:
1. scie-jump -> re-exec scie-pants Rust
2. scie-pants Rust -> re-exec scie-jump
3. scie-jump -> re-exec pants-dev

Doing the re-direct in scie-pants saves #3 which is likely ~300us. So, you could get away with it, but this is not a style choice or even a perf choice really. Those other cases all need binding commands to run and that's why they re-direct. You do not, you know what to re-exec and it needs no bindings setup at all.

jsirois:
> I'm not sure how/where to write tests for this, but it seems to do what I want when tested on the Pants repo.

You just reviewed the ~same testing needs here:
https://github.com/pantsbuild/scie-pants/pull/77/files#diff-7ff30d851f930debdd0596654b3cb5b82ed6bf7b19db396907dd2a7564514ffcR931

That clones the pants repo quickly at a fixed sha and builds pants and caches it all (CI too). With that - a Pants repo - you have what you need to verify `scie-pants` works when run with that clone as its CWD.

package/scie-pants.lift.json
jsirois: Here's a rule of thumb - perhaps even a hard rule, I can't see exceptions yet:

If you define a named "commands" entry that uses 0 placeholders; that's incorrect unless your scie is a BusyBox (it has no default "" command).

jsirois: So, you did use a placeholder here ... but you had to fill that in yourself in the default command; thus the wasteful bit to use an extra re-exec than needed for dispatch.

kaos: Makes total sense, thank you.

Spent 80% of the time spent on this PR on reading code to build my mental map of the terrain here (still incomplete ;)

kaos:
> > I'm not sure how/where to write tests for this, but it seems to do what I want when tested on the Pants repo.
> 
> You just reviewed the ~same testing needs here: https://github.com/pantsbuild/scie-pants/pull/77/files#diff-7ff30d851f930debdd0596654b3cb5b82ed6bf7b19db396907dd2a7564514ffcR931
> 
> That clones the pants repo quickly at a fixed sha and builds pants and caches it all (CI too). With that - a Pants repo - you have what you need to verify `scie-pants` works when run with that clone as its CWD.

Yup, I recognized it when I saw it--thanks for not assuming I did :)

kaos:
Hmm... locally this works for me 🤔  -- not sure yet what's different here.

kaos:
> Hmm... locally this works for me 🤔 -- not sure yet what's different here.

Ah, it seems that the pants repo is not re-fetched and thus patched. So my new patch is never applied..

Is there any way to clear the caches?

jsirois (approved):
Thanks for working through this @kaos

PR #77, opened 2023-01-27
Add `pants_from_sources` support.

### LINKED ISSUE #30: Fold pants_from_sources into `scie-pants`
This seems doable and is more sane for end-users. One tool, one way to do things, no copying scripts around.

A current example of the `pants_from_sources` script is here: https://github.com/pantsbuild/example-python/blob/main/pants_from_sources

jsirois:
Closes #30


jsirois:
1 extra commit coming to add CI caching of `~/.cache/scie-pants/dev` since its so slow to build the Pants used for the pants_from_sources tests - but safe to review.

src/main.rs
```diff
@@ -183,6 +184,36 @@ fn get_pants_process() -> Result<Process> {
...
+        pants_repo_location
+            .join("src")
+            .join("python")
+            .join("pants")
+            .join("VERSION"),
+    )?;
+
+    // When running pants from sources you are likely to be modifying those sources, so you won't
+    // want pantsd running.  You can override this by setting ENABLE_PANTSD=true.
+    let enable_pantsd = env::var_os("ENABLE_PANTSD").unwrap_or_else(|| "false".into());
```
jsirois, line 203: This is actually super weird. Why not just use `PANTS_PANTSD`?! This comment is copied from the script, but of course no longer visible to the end user. I'll add support from reading from `PANTS_PANTSD` which is the obvious thing to do for an end user (or use the flag, which will work as well).

jsirois: I guess this was maybe some - misguided imo - attempt at letting users avoid the double dribble of PANTS_PANTSD? That is a real way to configure Pants though that is well documented - etc. Not to mention this is advanced use anyhow - changing the default for a pants from source run - so if any user can handle the awkward double dribble, its thie user in this mode's audience.

CHANGES.md
```diff
@@ -1,5 +1,23 @@
...
+This release adds support for running Pants from a local Pants clone. This is useful for testing out
+unreleased Pants changes.
+
+This feature used to be provided by a bespoke `pants_from_sources` script copied around to various
+repositories; an example of which is [here](
+https://github.com/pantsbuild/example-python/blob/1b38d08821865e3756024950bc000bdbd0161b95/pants_from_sources).
+
+There are two ways to activate this mode:
+1. Execute `pants` with the `PANTS_SOURCE` environment variable set as the path to the Pants repo
+   whose Pants code you'd like to run against you repo.
```
kaos, line 14:
```diff
-against you repo.
+against your repo.
```

README.md
```diff
@@ -63,6 +63,13 @@ provides the following:
   If you run `scie-pants` in a directory where Pants is not already set up, it will prompt you, and
   you can let it set up the latest Pants stable version for your project.
 
++ Built-in [`pants_from_sources`](
+  https://github.com/pantsbuild/example-python/blob/1b38d08821865e3756024950bc000bdbd0161b95/pants_from_sources)
+  support. You can either execute `scie-pants` with `PANTS_SOURCE` set to the path of a local clone
+  of the [Pants](https://github.com/pantsbuild/pants) repo or else copy, link or symlink you
```
kaos, line 69:
```diff
-symlink you
+symlink your
```

kaos (approved):

kaos:
Ah, I see.. is this also closer in line with how you'd prefer to solve #33 ?

jsirois:
> Ah, I see.. is this also closer in line with how you'd prefer to solve https://github.com/pantsbuild/scie-pants/issues/33 ?

This really bears no relation to #33 in that it is completely driven by preexisting requirements. The `pants_from_sources` scripts totally constrained what I could do here. There were 0 choices for me to make afaict.

#33 can be solved by using any unambiguous marker. If and only if it's decided the marker lives in pants.toml / is an env var, then the soft constraint of not grabbing more subsystem name land is to re-use BOOTSTRAP which was grabbed by `PANTS_BOOTSTRAP_TOOLS` before scie-pants was even an idea.

CHANGES.md
jsirois: Fixed.

README.md
jsirois: Fixed.

src/main.rs
jsirois: Fixed.

kaos:
Sorry I was being very vague in my phrasing of that question. Having just reviewed the implementation I was thinking of the implementation--regardless of how we detect that we want to delegate to a `./pants` bootstrap script--in `main.rs` I had made another lift-entry for `pants-dev`, where as here you made another get process function altogether. I see those as two different ways of solving this, and both could work for #33 as well, I think. So, seeing this method here now, in contrast to what I cooked up in #75 I was wondering if this approach may be preferable to having that extra lift-entry?

But, we can defer this topic to the review of #75. :)

jsirois:
> But, we can defer this topic to the review of https://github.com/pantsbuild/scie-pants/pull/75. :)

I answered there, but the key difference in "commands" delegation is the need for "bindings" (~install setup steps). You do not have that need.

src/main.rs
benjyw: I suspect this wasn't a conscious attempt to avoid the double dribble, but that the script author didn't think to use the regular option env var for this.

benjyw:
Sorry I'm late to the party, but it occurs to me that `PANTS_SOURCE` lives in the `PANTS_` namespace of Pants options. E.g., if we add a global option named `--source`, this would stomp it. 

In fact, we already have a "source" subsystem, for source root configuration, so `PANTS_SOURCE_*` is occupied. This doesn't collide with `PANTS_SOURCE`, but I wonder if it's close enough to be confusing?

Should this be `_PANTS_SOURCE`? Or maybe this isn't a real problem in practice.

PS I like the env var a lot more than pants_from_sources and will be using it exclusively from now on.

jsirois:
Yeah, the current choice matches status quo, but I do agree status quo is confusing. I'm on a mission to do a narrow thing: replace ./pants and not break people. If you or others want to follow up and do other things, that's great.

benjyw:
Uurgh, didn't realize we were using `PANTS_SOURCE` inside `pants_from_sources` today.

I'll think about whether this is worth changing.

PR #79, opened 2023-01-27
Fix Release Announcement.

jsirois:
Previously it triggered 3x as each release shard finished. Now it triggers once at the end of the GH Actions release activities.


jsirois:
The triple announce this fixes can be seen here: https://pantsbuild.slack.com/archives/C18RRR4JK/p1674808299671799

kaos (approved):

PR #84, opened 2023-01-29
Fix pants_from_sources argv forwarding.

### LINKED ISSUE #83: pants_from_sources mode duplicates command-line arguments when invoking pants
The `Process::args` field appears to be intended to store any arguments used to invoke the Pants binary but not the user's command-line arguments (e.g. goals and targets). The user's arguments are already appended to the final command line by https://github.com/pantsbuild/scie-pants/blob/a3d85cfe6b1b35bf662a1f09db413f4abfa37781/src/main.rs#L59. But when setting up the `Process` for pants_from_sources mode the command-line arguments are also appended by https://github.com/pantsbuild/scie-pants/blob/a3d85cfe6b1b35bf662a1f09db413f4abfa37781/src/main.rs#L191. It should be noted that `get_pants_process` does not set `Process:args`.

jsirois:
Fixes #83

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Release Notes
 
+## 0.3.1
+
+This release fixes the Pants from sources feature added in 0.3.0 to forward command line arguments
+to the Pants run from sources correctly. Previously the argument list passed was doubled.
+
 ## 0.3.0
 
 This release adds support for running Pants from a local Pants clone. This is useful for testing out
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -1,5 +1,10 @@
 # Release Notes
 
+## 0.3.1
```
jsirois, line 3: N.B.: I'm not releasing, just adding to release notes ahead of time. I'd like to wait for the brew tap release integration to land before releasing next to test that out.

kaos (approved):

PR #88, opened 2023-01-29
Fix github-token value typo.

jsirois:
The terminology is PAT (Personal Access Token) and the secret is configured as such.


kaos:
for posterity: the idea was to call it TAP_TOKEN as it's scoped for the pantsbuild/homebrew-tap repo, and to avoid squatting on a very generic token name. was mis-named in the secrets.

PR #97, opened 2023-02-03
Prepare the 0.4.0 release.

jsirois:


kaos:
Feels like we'd want to include #98 in this release as well..?

jsirois:
If you'd like, sure. I'll hold off merge but keep the PR open.

In my accounting that issue adds a new feature not supported by `./pants` today at all; so it needs not block this feature. You're, presumably, anticipating a bug report by a fellow Pants dev who guesses this new feature works already.

kaos (approved):
You're right. #98 is the opt-out for the #33 toggle, being closely related but a separate feature.

It would work (at least better) in Pants if it weren't for the broken (i.e. not released) explorer backend.

If you don't mind frequent releases, no need to hold this (might even give us an opportunity to try out the brew tap release integration)

jsirois:
> If you don't mind frequent releases, no need to hold this (might even give us an opportunity to try out the brew tap release integration)

Exactly. I'm both a fan of frequent releases from main with no branching (~never break people, ~never advance the major version) and it would be great to try out the brew integration again.

Ok, I'll proceed.

jsirois:
And the tap-integration worked @kaos. Thanks again for that.

PR #100, opened 2023-02-03
Set `_PANTS_OVERRIDE_VERSION` when delegate_bootstrap has been overridden.

### LINKED ISSUE #98: Set `_PANTS_OVERRIDE_VERSION` when punting the `delegate_bootstrap`
Aha, so I get the feeling the correct thing to do would be to provide `_PANTS_VERSION_OVERRIDE={pants_version}` if we have `delegate_bootstrap` and then continue to run the normal `pants` flow, as that would "cancel" the delegation to `./pants` and provide the "fix" to work with a released version of Pants in the pants repo.

I can put up a PR for this later tonight.

_Originally posted by @kaos in https://github.com/pantsbuild/scie-pants/issues/33#issuecomment-1416076160_

kaos:
Special case when running in the Pants source repo and `delegate_bootstrap` is enabled--when a specific version of Pants is requested (either via `pants_version` in the `pants.toml` file or using env vars `PANTS_VERSION` or `PANTS_SHA`) bootstrap using that version of Pants rather than delegating to `./pants`.

Closes #98

### DIFF
--- a/package/src/main.rs
+++ b/package/src/main.rs
@@ -131,10 +131,7 @@ macro_rules! build_step {
 
 macro_rules! integration_test {
     ($msg:expr $(,)?) => {
-        log!(Color::Magenta, ">> {}", $msg);
-    };
-    ($msg:expr, $($arg:tt)*) => {
-        log!(Color::Magenta, ">> {}", format!($msg, $($arg)*));
+        log!(Color::Magenta, ">> {}", format!($msg));
     };
 }
 
@@ -1083,18 +1080,22 @@ index b70ae75..271706a 100644
             rename(&venv_root_tmp.into_path(), &pants_2_14_1_venv_dir)?;
         }
 
+        let decode_output = |output: Vec<u8>| {
+            let res = String::from_utf8(output).map_err(|e| {
+                Code::FAILURE.with_message(format!("Failed to decode Pants output: {e}"))
+            })?;
+            Ok(res)
+        };
         let test_pants_from_sources = |command: &mut Command, expected_messages: Vec<&str>| {
             let result = execute(command.arg("-V").stderr(Stdio::piped()))?;
-            let stderr = String::from_utf8(result.stderr).map_err(|e| {
-                Code::FAILURE.with_message(format!("Failed to decode Pants stderr: {e}"))
-            })?;
+            let stderr = decode_output(result.stderr.clone())?;
             for expected_message in expected_messages {
                 assert!(
                     stderr.contains(expected_message),
                     "STDERR did not contain '{expected_message}':\n{stderr}"
                 );
             }
-            Ok(stderr)
+            Ok(result)
         };
 
         test_pants_from_sources(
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -179,6 +179,9 @@ fn get_pants_process() -> Result<Process> {
         ));
     }
     if let Some(version) = pants_version {
+        if delegate_bootstrap {
+            env.push(("_PANTS_OVERRIDE_VERSION".into(), version.clone().into()));
```
jsirois, line 183: I think it would be good to follow up on the ick by adding comments to this env var over in Pants (as well as the build root override env var) to point out they are being used externally here; i.e.: don't yank or rename without consultation / coordination.

jsirois (approved):
Excellent, thank you.

jsirois:
@kaos if you want to try your hand at a release prep / release I'll let you do so; otherwise I'm happy to pick that up presently.

src/main.rs
kaos: Good call. Will do. 👍🏽

kaos:
> @kaos if you want to try your hand at a release prep / release I'll let you do so; otherwise I'm happy to pick that up presently.

Would be happy to; guessing it is rather straight forward. Today however I'll let you run with it, as it's our anniversary and planning on going out for dinner (16 years married to my wife today 🎉 time flies)

kaos:
hmm... is it a flaky test to look for the scheduler init string? 😬

package/src/main.rs
```diff
@@ -1127,30 +1123,43 @@ index b70ae75..271706a 100644
...
+                .env("PANTS_VERSION", "2.16.0.dev5")
+                .env(
+                    "PANTS_BACKEND_PACKAGES",
+                    "-[\
+                    'internal_plugins.test_lockfile_fixtures',\
+                    'pants.backend.explorer',\
+                    ]",
+                )
+                .current_dir(&pants_2_14_1_clone_dir),
+            vec!["[INFO] Scheduler initialized."],
```
jsirois, line 1157: Yeah, this could be flaky maybe? It relies on the conditions dictating pantsd must restart. Here the underlying codebase is identical and just the ~PANTS_VERSION env var has changed which may not be enough to change the pantsd fingerprint which is the only thing guaranteed to cause a pantsd restart. The rest are due to exceeding memory thresholds.

I'm definitely a bit muddy on the above, but an alternative is to do a more direct test. Have the Command capture stdout and re-jigger the test closure so you can get at stdout / the process result and just verify the printed version is 2.16.0.dev5. Other ITs do this sort of thing.

jsirois: Right here for example, although with stderr:
https://github.com/pantsbuild/scie-pants/blob/7fc1471d6d0d66615fb0217459d5942cd161a799/package/src/main.rs#L819-L833

kaos: thanks, will circle back to that later :)

jsirois:
Aha, congratulations! Yeah, I'll hold off on release since there are issues anyhow.

kaos:
I've put the test lines in a block for readability on a few of the tests to try it out. If you agree with it, I can go over and apply that to all the integration tests. For me, it made it more readable where the different tests started/ended. (also hoping to reduce local variable pollution between tests, although some of the tests rely on it, so those would likely be left outside such a block in that case)

src/main.rs
kaos: pantsbuild/pants#18170

package/src/main.rs
```diff
@@ -1083,75 +1080,103 @@ index b70ae75..271706a 100644
...
+                    .current_dir(&pants_2_14_1_clone_dir),
+                vec![
+                    "The delegate_bootstrap mode is working.",
+                    "Pants from sources argv: -V.",
+                ],
+            )?;
+        }
+
+        let pants_release = "2.16.0.dev5";
+        integration_test!("Verify usage of a Pants {pants_release} on the pants repo.");
```
kaos, line 1151: Will fix on next push:

```suggestion
        integration_test!("Verify usage of Pants {pants_release} on the pants repo.");
```

package/src/main.rs
```diff
@@ -131,10 +131,7 @@ macro_rules! build_step {
 
 macro_rules! integration_test {
     ($msg:expr $(,)?) => {
-        log!(Color::Magenta, ">> {}", $msg);
-    };
-    ($msg:expr, $($arg:tt)*) => {
-        log!(Color::Magenta, ">> {}", format!($msg, $($arg)*));
+        log!(Color::Magenta, ">> {}", format!($msg));
```
jsirois, line 134: You should be able to drop the format! when you circle back and just pass $msg directly.

jsirois (approved):

package/src/main.rs
kaos: I'm not so sure, as the reason I touched this part was due to (from my console output):
```
>> Verify usage of a Pants {pants_release} on the pants repo.
```

so the arg-less version of the macro needs the `format!` it seems.

kaos: and I dropped the arg version as after the Rust upgrade, it seems embedding the args are the new way to go about formatting.. :)

kaos: Yep, confirmed. Still need `format!`.

PR #108, opened 2023-02-07
Fix `.pants.bootstrap` handling.

### LINKED ISSUE #107: `exported` array setting in `.pants.boostrap` parsed differently by `pants` than by `./pants`
In our `.pants.bootstrap`, we have:
```
export PANTS_DOCKER_TOOLS="+['aws-oidc', 'open']"
```
(It's a conditional `export`, otherwise we'd put it in `pants.toml`)

When running `./pants`, this `export` works to append the 2 values to the end of the config specified in `pants.toml`. When running the `pants` binary, the setting instead seems to overwrite the existing value with the escaped string contents of the env var, resulting in the error:
```
pants.core.util_rules.system_binaries.BinaryNotFoundError: Cannot find `'+['\''aws-oidc'\'', '\''open'\'']'` on `['/System/Cryptexes/App/usr/bin', '/Users/dan.moran/.asdf/shims', '/Users/dan.moran/.cargo/bin', '/Users/dan.moran/.go/bin', '/Users/dan.moran/bat-extras/bin', '/bin', '/opt/homebrew/bin', '/sbin', '/usr/bin', '/usr/local/MacGPG2/bin', '/usr/local/bin', '/usr/local/go/bin', '/usr/local/opt/asdf/libexec/bin', '/usr/local/opt/python3/libexec/bin', '/usr/local/sbin', '/usr/sbin']`. Please ensure that it is installed so that Pants can use docker.
```

jsirois:
A re-exec approach is now used to ensure pants runs in a bash subprocess
for complete fidelity with the `./pants` bash script case.

Fixes #107


src/main.rs
```diff
@@ -86,15 +85,66 @@ fn env_version(env_var_name: &str) -> Result<Option<String>> {
...
+
+    fn into_process(
+        self,
+        scie: String,
+        build_root: Option<PathBuf>,
+        env: Vec<(OsString, OsString)>,
+    ) -> Result<Process> {
+        Ok(match build_root.map(|br| br.join(".pants.bootstrap")) {
+            Some(pants_bootstrap) if self != Self::BootstrapTools && pants_bootstrap.is_file() => {
+                Process {
```
jsirois, line 124: This is obviously a much more robust way to do things. I read https://github.com/pantsbuild/pants/issues/16351#issuecomment-1328250334 / the associated discussion uncritically when I 1st implemented this and thought bash variables (as well as env vars) needed to be captured; thus the prior approach. Only exported env vars were making it to the pants console script from the `./pants` wrapper script though; so this is all that is needed.

danxmoran (approved):
Makes sense to me, thanks for the quick fix!

kaos (approved):

PR #112, opened 2023-02-08
Improve Pants help output / CLI fidelity.

### LINKED ISSUE #111: Set a friendlier `PANTS_BIN_NAME`
Currently the absolute path of scie-pants is used. This is correct, but in the common case its verbose and not what the user actually typed. Worse, Pants uses this value in ways it probably should not and gets absolute paths written into files where they do not belong.

See https://github.com/pantsbuild/pants/issues/18202, but the idea is to just use the basename of scie-pants if it is on the PATH and otherwise relativize it if it is within the `BUILDROOT` tree finally falling back to the current absolute path in all other cases.

jsirois:
Previously scie-pants over-rode Pants help defaulting to `./pants`
in help text by passing `--pants-bin-name` as the absolute path of
scie-pants. Now scie-pants passes the exact argv0 string it was
invoked with and it does so using the `PANTS_BIN_NAME` env var instead
of `--pants-bin-name` to support users over-riding this.

Closes #111

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Release Notes
 
+## 0.5.0
+
+This release improves `scie-pants` operation with Pants help by ensuring the command line you used
+to invoke Pants is accurately reflected in the help information Pants presents back to you.
+
 ## 0.4.2
 
 This release fixes `.pants.bootstrap` handling to robustly mimic handling by the `./pants` script.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -585,7 +585,7 @@ dependencies = [
 
 [[package]]
 name = "scie-pants"
-version = "0.4.2"
+version = "0.5.0"
 dependencies = [
  "anyhow",
  "dirs",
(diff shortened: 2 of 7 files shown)


kaos (approved):
Out of curiosity what changes do you classify as a minor bump vs a point bump? (or major bump for that matter, but that one is easier I think--for very breaking changes..)

jsirois:
I'm hoping not to be idiosyncratic and just follow semver. This one either is a bug fix (you couldn't set PANTS_BIN_NAME from the outside before) or a new feature (the help strings you get are better now / reflect what you typed). I leaned towards the latter, but if you think this is incorrect semver I want to correct that and use a patch bump instead; so let me know.

kaos:
Not incorrect at all, just figuring that if I'll do another release in the future, to make the proper bump ;) 

Minor bump for feature change, point bump for bugfix. Makes sense to me.

Also, that you couldn't set the bin name before might as well be a missing feature (as it wasn't "broken" per se).

PR #118, opened 2023-02-11
Break up package tool monolith.

jsirois:
Break out logical modules and individual test functions.


jsirois:
At ~1600 lines the package main was getting out of hand. This is somewhat unreviewable, but I've put involved folks on the review. Suffice it to say this is all just code motion with no changes otherwise. The most interesting bit as a contributor will be the breaking out of a test module, which is where you interact with all this when contributing fixes and features to the main scie-pants binary.

stuhood (approved):

PR #119, opened 2023-02-12
Simplify error handling with anyhow.

jsirois:
This was already used by scie-pants but the package crate is now converted as well.


benjyw (approved):
That's a neat library with a fun name.

jsirois:
> That's a neat library with a fun name.

I'm happy when I can use libraries from ~3-4 folks and David Tolnay is one of them. Good stuff.

PR #125, opened 2023-02-20
Fix `pants run --debug-adapter`.

jsirois:
This fixes `scie-pants` to work with `pants run --debug-adapter` by not
setting an empty `PANTS_DEBUGPY_VERSION` env var.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Release Notes
 
+## 0.5.2
+
+This release fixes `scie-pants` to interoperate with `pants run --debug-adapter`. Previously, if
+there was no custom `[debugpy] version` configured, `scie-pants` would cause Pants to error by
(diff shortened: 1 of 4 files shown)


jsirois:
I found this trying to debug a user's `--debug-adapater` issue with ns-packages in Slack.

thejcannon:
For my own understanding, we care about this _at all_ because of the case where someone is trying to debug Pants itself right? Otherwise, this env var would be like any other Pants-option-from-an-env-var?

jsirois:
> For my own understanding, we care about this at all because of the case where someone is trying to debug Pants itself right? 

Correct. One small feature above and beyond `PANTS_DEBUG=1 ./pants` is that `PANTS_DEBUG=1 pants` will honor your custom `debugpy` configuration.

thejcannon (approved):

thejcannon:
Well that's awesome :smile:

PR #136, opened 2023-03-06
Do not change directory when running pants from sources.

kaos:


src/main.rs
```diff
@@ -280,15 +280,17 @@ fn get_pants_from_sources_process(pants_repo_location: PathBuf) -> Result<Proces
         .or_else(|| env::var_os("PANTS_PANTSD"))
         .unwrap_or_else(|| "false".into());
 
+    let build_root = BuildRoot::find(None)?;
     let env = vec![
         ("PANTS_VERSION".into(), version.trim().into()),
         ("PANTS_PANTSD".into(), enable_pantsd),
+        (
+            "PANTS_BUILDROOT_OVERRIDE".into(),
+            build_root.as_os_str().to_os_string(),
```
kaos, line 289: This didn't seem needed, at the same time I think it would be better to not get rid of at least the `BuildRoot::find(None)?` call to catch an unknown build root early, and then we might as well set this too..?

jsirois: It's only not needed for Pants 2.16.0.dev7 and newer: https://github.com/pantsbuild/pants/pull/18170

If you want to run pants from older sources it's needed if the non Pants repo your working in doesn't have a pants script checked in.

jsirois (approved):

jsirois (commented):
Um, you need to bump the version in the root Cargo.toml.

src/main.rs
kaos: Oh... heh 🤦🏽

PR #156, opened 2023-04-09
Improve error messaging on bad Pants version

### LINKED ISSUE #155: scie-pants unhelpful error on ambiguous Pants version tag
For example, setup pants.toml with version 2.16.0

```
$ pants
Failed to fetch https://binaries.pantsbuild.org/tags/pantsbuild.pants/release_2.16.0: [22] HTTP response code said error (The requested URL returned error: 404)
Install failed: list indices must be integers or slices, not str
More information can be found in the log at: /Users/navneethjayendran/Library/Caches/nce/142382d6e3ac0960e7d4f81b9df9de9559b7ac698db13d20cbd0e4aa57f68c1d/bindings/logs/configure.log

Error: Isolates your Pants from the elements.

<default>: Detects the current Pants installation and launches it.
bootstrap-tools: Introspection tools for the Pants bootstrap process.
pants: Runs a hermetic Pants installation.
pants-debug: Runs a hermetic Pants installation with a debug server for debugging Pants code.
update: Update scie-pants.

You can select a boot command by passing it as the 1st argument or else by setting the SCIE_BOOT environment variable.

ERROR: Failed to establish atomic directory /Users/navneethjayendran/Library/Caches/nce/142382d6e3ac0960e7d4f81b9df9de9559b7ac698db13d20cbd0e4aa57f68c1d/locks/configure-0cd3605ca31ca83f339db1732ae6d5e06250818e543746695e31695fcc251cad. Population of work directory failed: Boot binding command failed: exit status: 1
```

This seems to happen because the Github API, instead of 404'ing, will return a list of potential matches when the tag is a prefix of existing tags but doesn't exactly match any particular tag. It's a slightly confusing error that could be clarified.

NavneethJayendran:
Resolves #155

Handle the case where the Github API returns multiple results due to an inexact release tag match
by raising a new `TagNotFoundError` which also links the Pants releases page. Also wrap the typical
`CalledProcessError` with this new exception.

### DIFF
--- a/tools/src/scie_pants/pants_version.py
+++ b/tools/src/scie_pants/pants_version.py
@@ -8,6 +8,7 @@ import json
 import logging
 import os
 import urllib.parse
+from typing import Any, Dict, List
 from dataclasses import dataclass
 from pathlib import Path
 from subprocess import CalledProcessError
@@ -42,6 +43,8 @@ class ResolveInfo:
         )
         return f"--python-repos-{option_name}={operator}['{self.find_links}']"
 
+class TagNotFoundError(Exception):
+    pass
 
 def determine_find_links(
     ptex: Ptex,
@@ -117,12 +120,22 @@ def determine_tag_version(
         github_api_url = (
             f"https://api.github.com/repos/pantsbuild/pants/git/refs/tags/{urllib.parse.quote(tag)}"
         )
+        github_releases_url="https://github.com/pantsbuild/pants/releases"
         headers = (
             {"Authorization": f"Bearer {github_api_bearer_token}"}
             if github_api_bearer_token
             else {}
         )
-        github_api_tag_url = ptex.fetch_json(github_api_url, **headers)["object"]["url"]
+        github_api_response: Dict[str, Any]
+        try:
+            github_api_response = ptex.fetch_json(github_api_url, **headers)
+            # If the response is a list, multiple matches were found, which means the supplied tag
+            # is not an exact match against any tag
+            if isinstance(github_api_response, list):
+                raise TagNotFoundError(f"Could not find Pants tag {tag} in {github_releases_url}")
+        except CalledProcessError as e:
+            raise TagNotFoundError(f"Could not find Pants tag {tag} in {github_releases_url}: {e}")
+        github_api_tag_url = github_api_response["object"]["url"]
         commit_sha = ptex.fetch_json(github_api_tag_url, **headers)["object"]["sha"]
 
     return determine_find_links(


tools/src/scie_pants/pants_version.py
```diff
@@ -117,12 +120,22 @@ def determine_tag_version(
...
             if github_api_bearer_token
             else {}
         )
-        github_api_tag_url = ptex.fetch_json(github_api_url, **headers)["object"]["url"]
+        github_api_response: Dict[str, Any]
+        try:
+            github_api_response = ptex.fetch_json(github_api_url, **headers)
+            # If the response is a list, multiple matches were found, which means the supplied tag
+            # is not an exact match against any tag
+            if isinstance(github_api_response, list):
```
jsirois, line 134: It seems like it would make a but more sense to keep the comment mostly unchanged but change the test to `if not isinstance(github_api_response, dict)` since that is what is used / relied upon on line 138 below. As it stands, if the API returned a string a bool or a number, the same style of confusing exception as this fixes would be raised.

jsirois (commented):
Thanks @NavneethJayendran. Are you up for adding a test?

That might go here as a call to a new `test_bad_pants_version` test function you'd add lower down: https://github.com/pantsbuild/scie-pants/blob/b92a07685822a019c1682e23ca3ec100de07934d/package/src/test.rs#L98

NavneethJayendran:
@jsirois Sorry for the long response time. I adjusted the check to look for a dict specifically. @AIGeneratedUsername I added a test case for this error message as you suggested.

jsirois:
@NavneethJayendran thanks for picking this back up. It looks like there are test issue to work out still. I'm going to add a few folks as reviewers since I've stepped away from Pants-related development. They should be able to help you take this home or else recommend further reviewers.

package/src/test.rs
```diff
@@ -38,7 +40,7 @@ fn decode_output(output: Vec<u8>) -> Result<String> {
 }
 
 fn assert_stderr_output(command: &mut Command, expected_messages: Vec<&str>) -> Output {
-    let output = execute(command.stderr(Stdio::piped())).unwrap();
+    let output = execute_no_error(command.stderr(Stdio::piped()));
```
kaos, line 43: I worry that this does no longer make the same assertion about the sub process' exit code as before for all the unrelated tests using this method now.

tools/src/scie_pants/pants_version.py
```diff
@@ -117,12 +121,22 @@ def determine_tag_version(
         github_api_url = (
             f"https://api.github.com/repos/pantsbuild/pants/git/refs/tags/{urllib.parse.quote(tag)}"
         )
+        github_releases_url = "https://github.com/pantsbuild/pants/releases"
```
thejcannon, line 124: So we're actually about to switch to a GitHub Release-based approach for any version >=2.

If you're able to sit a smidge, this code might look different when we make that switch.

NavneethJayendran: Let me know when that happens and I can revisit.

package/src/test.rs
huonw: As of #296, callers of `assert_stderr_output` can indicate whether the command should either succeed or fail.

tools/src/scie_pants/pants_version.py
huonw: Thanks for waiting! The switch has happened now.

For version strings that look new (after `PANTS_PEX_GITHUB_RELEASE_VERSION`), this code path isn't used, as the function returns early. For older ones, definitely still worth considering improvements'; thanks!

For those new ones, Pants is then downloaded from GitHub directly, with some error handling about it not existing (but feel free to fine-tune if you have ideas!), in https://github.com/pantsbuild/scie-pants/blob/91477f6394e8eed048e093538b4a4c299514e44f/tools/src/scie_pants/install_pants.py#L88-L116

benjyw:
Hi @NavneethJayendran ! Just checking on the status of this?

PR #172, opened 2023-05-08
Use the pants native-client if it is present in the distribution.

stuhood:
This change uses the `native-client` binary included in `pantsbuild/pants` wheels by https://github.com/pantsbuild/pants/pull/18957 (and further adjusted in https://github.com/pantsbuild/pants/pull/19010) to launch `pants`. See that change for the performance impact.

### DIFF
--- a/package/scie-pants.lift.json
+++ b/package/scie-pants.lift.json
@@ -92,21 +92,10 @@
             "description": "Runs a hermetic Pants installation.",
             "env": {
               "=PANTS_VERSION": "{scie.bindings.configure:PANTS_VERSION}",
-              "PANTS_BUILDROOT_OVERRIDE": "{scie.bindings.configure:PANTS_BUILDROOT_OVERRIDE}"
-            },
-            "exe": "{scie.bindings.install:VIRTUAL_ENV}/bin/pants",
-            "args": [
-              "{scie.bindings.configure:PANTS_SHA_FIND_LINKS}"
-            ]
-          },
-          "pants-maybe-native-client": {
-            "description": "Runs a hermetic Pants installation via the native client (falling back to the legacy client if the native client is not available).",
-            "env": {
-              "=PANTS_VERSION": "{scie.bindings.configure:PANTS_VERSION}",
               "PANTS_BUILDROOT_OVERRIDE": "{scie.bindings.configure:PANTS_BUILDROOT_OVERRIDE}",
-              "=_PANTS_SERVER_EXE": "{scie.bindings.install:VIRTUAL_ENV}/bin/pants"
+              "=_PANTS_SERVER_EXE": "{scie.bindings.install:PANTS_SERVER_EXE}"
             },
-            "exe": "{scie.bindings.install:MAYBE_NATIVE_CLIENT_BINARY}",
+            "exe": "{scie.bindings.install:PANTS_CLIENT_EXE}",
             "args": [
               "{scie.bindings.configure:PANTS_SHA_FIND_LINKS}"
             ]
--- a/package/scie-pants.lift.json
+++ b/package/scie-pants.lift.json
@@ -99,6 +99,18 @@
               "{scie.bindings.configure:PANTS_SHA_FIND_LINKS}"
             ]
           },
+          "pants-maybe-native-client": {
+            "description": "Runs a hermetic Pants installation via the native client (falling back to the legacy client if the native client is not available).",
+            "env": {
+              "=PANTS_VERSION": "{scie.bindings.configure:PANTS_VERSION}",
+              "PANTS_BUILDROOT_OVERRIDE": "{scie.bindings.configure:PANTS_BUILDROOT_OVERRIDE}",
+              "=_PANTS_SERVER_EXE": "{scie.bindings.install:VIRTUAL_ENV}/bin/pants"
+            },
+            "exe": "{scie.bindings.install:MAYBE_NATIVE_CLIENT_BINARY}",
+            "args": [
+              "{scie.bindings.configure:PANTS_SHA_FIND_LINKS}"
+            ]
+          },
           "pants-debug": {
             "description": "Runs a hermetic Pants installation with a debug server for debugging Pants code.",
             "env": {
--- a/tools/src/scie_pants/install_pants.py
+++ b/tools/src/scie_pants/install_pants.py
@@ -10,6 +10,7 @@ import sys
 from argparse import ArgumentParser
 from pathlib import Path
 from typing import Iterable, NoReturn
+from glob import glob
 
 from packaging.version import Version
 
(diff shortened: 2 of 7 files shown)


package/scie-pants.lift.json
```diff
@@ -99,6 +99,18 @@
...
             ]
           },
+          "pants-maybe-native-client": {
+            "description": "Runs a hermetic Pants installation via the native client (falling back to the legacy client if the native client is not available).",
+            "env": {
+              "=PANTS_VERSION": "{scie.bindings.configure:PANTS_VERSION}",
+              "PANTS_BUILDROOT_OVERRIDE": "{scie.bindings.configure:PANTS_BUILDROOT_OVERRIDE}",
+              "=_PANTS_SERVER_EXE": "{scie.bindings.install:VIRTUAL_ENV}/bin/pants"
+            },
+            "exe": "{scie.bindings.install:MAYBE_NATIVE_CLIENT_BINARY}",
```
jsirois, line 109: Does the native client now handle starting pantsd when it's not already up?

stuhood: Yea: see https://github.com/pantsbuild/pants/pull/18957.

tools/src/scie_pants/install_pants.py
```diff
@@ -116,8 +117,13 @@ def main() -> NoReturn:
     )
     info(f"New virtual environment successfully created at {venv_dir}.")
 
+    # Added in https://github.com/pantsbuild/pants/commit/558d843549204bbe49c351d00cdf23402da262c1
+    native_client_binaries = glob(str(venv_dir / "lib/python*/site-packages/pants/bin/native_client"))
+    maybe_native_client_binary = native_client_binaries[0] if len(native_client_binaries) == 1 else venv_dir / "bin" / "pants"
```
jsirois, line 122: Hrm, it looks like CI is running isort & fmt but not in check mode - this is >100 cols: https://github.com/pantsbuild/scie-pants/actions/runs/4941570260/jobs/8834300615?pr=172#step:10:400

package/scie-pants.lift.json
```diff
@@ -99,6 +99,18 @@
               "{scie.bindings.configure:PANTS_SHA_FIND_LINKS}"
             ]
           },
+          "pants-maybe-native-client": {
+            "description": "Runs a hermetic Pants installation via the native client (falling back to the legacy client if the native client is not available).",
+            "env": {
+              "=PANTS_VERSION": "{scie.bindings.configure:PANTS_VERSION}",
+              "PANTS_BUILDROOT_OVERRIDE": "{scie.bindings.configure:PANTS_BUILDROOT_OVERRIDE}",
+              "=_PANTS_SERVER_EXE": "{scie.bindings.install:VIRTUAL_ENV}/bin/pants"
```
jsirois, line 107: It seems like keeping the full value opaque and dried up in an install binding would be better. You calculate it there as well.

If the native client handled `PANTS_NO_NATIVE_CLIENT` processing as mentioned above, then the whole thing could just be handled by something like this change to the existing "pants" command:
```json
"pants": {
  "description": "Runs a hermetic Pants installation.",
  "env": {
    "=PANTS_VERSION": "{scie.bindings.configure:PANTS_VERSION}",
    "PANTS_BUILDROOT_OVERRIDE": "{scie.bindings.configure:PANTS_BUILDROOT_OVERRIDE}",
    "=_PANTS_SERVER_EXE": "{scie.bindings.install:PANTS_CONSOLE_SCRIPT}"
  },
  "exe": "{scie.bindings.install:PANTS_EXE}",
  "args": [
    "{scie.bindings.configure:PANTS_SHA_FIND_LINKS}"
  ]
},
```

package/scie-pants.lift.json
```diff
@@ -99,6 +99,18 @@
               "{scie.bindings.configure:PANTS_SHA_FIND_LINKS}"
             ]
           },
+          "pants-maybe-native-client": {
```
jsirois, line 102: I should have thought of this in the Pants-side PR, but if the native client binary could check `PANTS_NO_NATIVE_CLIENT` itself and immediately delegate then no new "pants-maybe-native-client" command would be needed. That would fit the spirit of the setting IIUC (almost no code executes to do that re-exec; so its not risky at all).

jsirois (approved):

tools/src/scie_pants/install_pants.py
jsirois: https://github.com/pantsbuild/scie-pants/pull/176

package/scie-pants.lift.json
stuhood: The concern that I had (not sure whether it was founded) was that the native-client might be completely non-functional on someone's machine, and that they might need to completely bypass invoking it.

jsirois: Yeah, that seems pretty unlikely to me!

stuhood: The I think that I was primarily thinking about dynamic linking... we attempt to build wheels on fairly old machines to avoid those issues, but. And maybe it's unlikely to experience _new_ dynamic linking issues that we haven't already experienced via the native-engine module?

But yea, I could be convinced to go back and implement it on that side instead.

jsirois: Any, fwiw, I build the scie-pants binary, jump and ptex as true static binaries using libmusl to avoid this sort of thing completely. Pants might do the same if you're truly worried. But, again, probably fruitless because the handoff to the engine will hit a bunch of links that will then fail.

stuhood:
Adjusted for https://github.com/pantsbuild/pants/pull/19010

jsirois (approved):

jsirois:
If you want to roll in the version bump & CHANGES entry, this could go straight to release.

package/scie-pants.lift.json
```diff
@@ -92,9 +92,10 @@
             "description": "Runs a hermetic Pants installation.",
             "env": {
               "=PANTS_VERSION": "{scie.bindings.configure:PANTS_VERSION}",
-              "PANTS_BUILDROOT_OVERRIDE": "{scie.bindings.configure:PANTS_BUILDROOT_OVERRIDE}"
+              "PANTS_BUILDROOT_OVERRIDE": "{scie.bindings.configure:PANTS_BUILDROOT_OVERRIDE}",
+              "=_PANTS_SERVER_EXE": "{scie.bindings.install:PANTS_SERVER_EXE}"
```
kaos, line 96: My brain would like these sorted.. no matter otherwise.

kaos (approved):

package/scie-pants.lift.json
stuhood: Technically they are sorted, due to the prefix `_`? But I don't think that it will begin to matter until/unless there are more of them.

PR #174, opened 2023-05-09
Fix PANTS_SHA install caching.

### LINKED ISSUE #173: Use of `PANTS_SHA` reexecutes some bootstrapping logic on each run
When `PANTS_SHA` is in use, a series of downloading lines flash by on every invoke.

jsirois:
Fixes #173

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@
 /.pants.d/
 /.pids/
 
+# Python bytecode
+__pycache__/
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## 0.6.1
+
+This release fixes `PANTS_SHA` support to properly cache the resulting Pants install.
+
 ## 0.6.0
 
 Support the `PANTS_TOML` environment variable for specifying a location other than the project's
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -614,7 +614,7 @@ dependencies = [
 
 [[package]]
 name = "scie-pants"
-version = "0.6.0"
+version = "0.6.1"
 dependencies = [
  "anyhow",
  "dirs",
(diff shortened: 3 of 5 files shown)


jsirois:
I don't know of a good way to test this. Manually tested of course, but ensuring the result is cached requires poking at internals or turning off the network.

kaos (approved):
> I don't know of a good way to test this.

In what context is the `PANTS_VERSION_PROMPT_SALT` available?

Two subsequent runs should have the same salt when this works, right?

jsirois:
> In what context is the PANTS_VERSION_PROMPT_SALT available?

Only when no Pants version is supplied (via pants.toml, PANTS_VERSION or PANTS_SHA)

> Two subsequent runs should have the same salt when this works, right?

They should have no PANTS_VERSION_PROMPT_SALT env var set at all, that was the bug. That's a very direct test of internals that doesn't do much though, I'm hoping for an IT, something that actually tests a back slide robust to details.

stuhood (approved):
Thanks!

----

Re: testing: yea, had the same realization for #172. I'm not sure what to do about that... maybe a Pants plugin which reports some details of the invoke? But it would be a big abstraction leak for any of that information to actually be exposed to plugins, so...

kaos:
> > In what context is the PANTS_VERSION_PROMPT_SALT available?
> 
> Only when no Pants version is supplied (via pants.toml, PANTS_VERSION or PANTS_SHA)
> 
> > Two subsequent runs should have the same salt when this works, right?
> 
> They should have no PANTS_VERSION_PROMPT_SALT env var set at all, that was the bug. That's a very direct test of internals that doesn't do much though. I'm hoping for an IT, something that actually tests a back slide robust to details.

Oh right. Yea, in my mind it merely changed the value. 

Guess an IT would have to capture network traffic or something like that maybe in order to detect the side-effect of using PANTS_SHA or not. Feels like a lot of work for a relatively small feature.

PR #190, opened 2023-05-25
Treat empty PANTS_VERSION and PANTS_SHA as if they were unset

huonw:
Overriding the pants version via the environment variables is convenient, and this PR makes it slightly more convenient by having `PANTS_VERSION= pants ...` treated the same as not setting it.

This means a script can do something like `for PANTS_VERSION in "" "2.15.0" ...;` to test against multiple versions, including whatever is natively configured in `pants.toml`, without having to conditionalise `PANTS_VERSION`'s existence entirely. For a concrete example, https://github.com/pantsbuild/pants/pull/19161 wants to be able to run a foreign repo's "native" pants version, but it's not trivial to do, due to this issue (the `""` part of the pants version matrix is commented out; obviously I could also have it set the env var conditionally, but that's less convenient).

Before this change, running `PANTS_VERSION= pants version` next to a `pants.toml` would prompt for initialising pants:

```
PANTS_VERSION= pants version                                                         
No Pants configuration was found at or above /Users/huon/projects/stellargraph/exoflare.
Would you like to configure /Users/huon/projects/stellargraph/exoflare as a Pants project? (Y/n):
```

After this change, it's the same as not setting `PANTS_VERSION`: use `pants.toml`'s setting if one exists, or prompt for initialisation, if there's no `pants.toml`.


src/main.rs
```diff
@@ -74,12 +74,14 @@ impl Process {
 }
 
 fn env_version(env_var_name: &str) -> Result<Option<String>> {
-    if let Some(raw_version) = env::var_os(env_var_name) {
+    let raw_version = env::var_os(env_var_name).unwrap_or(OsString::new());
+    if raw_version.len() == 0 {
+        // setting PANTS_VERSION= or PANTS_SHA= behaves the same as not setting them
```
jsirois, line 79: It would be great if this comment had a sibling test.

jsirois (approved):

src/main.rs
huonw: Ah, I had grepped for `#[test]` and didn't find anything, but (embarrassingly) just concluded "oh, I guess there's no tests", rather than trying to grep for `test` alone. I've found them now, will add one!

jsirois:
If you'd like to fold in steps 1-3 here: https://github.com/pantsbuild/scie-pants/blob/main/RELEASE.md#version-bump-and-changelog this can go straight to release when it lands on main.

huonw:
Done, thanks.

huonw:
Done: https://github.com/pantsbuild/scie-pants/actions/runs/5102247494

PR #192, opened 2023-05-26
Prepare the 0.8.0 release.

jsirois:


package/pbt.toml
```diff
@@ -1,57 +1,63 @@
...
-exe = "#{cpython:python}"
-args = [
-    "-m",
-    "pip"
-]
-
-[[lift.commands]]
-name = "python"
-exe = "#{cpython:python}"
+[lift]
```
jsirois, line 1: I'm sorry about this - the diff looks larger than it really is (just insertion of lines 7-11). I ran a neglected dos2unix on the file.

jsirois:
Hrm. It appears GitHub releases JSON for PBS is missing the "name" field most of the time - but not always - only on macOS 11 x86_64 ... which of course makes no sense unless a bad edge cache is between those macs and GH release data? I may try flipping to our self-hosted x86_64 mac as a debug step.

jsirois:
Hrm, yeah - switching to our self-hosted mac for the mac x86_64 shard seems to do it. 1 more change coming to upgrade all uses of ubuntu-20.04 to ubuntu-22.04 and that will tickle the new mac shard again to make sure the success is not a fluke.

.github/workflows/ci.yml
```diff
@@ -52,14 +52,14 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ${{ env.SCIE_PANTS_DEV_CACHE }}
-          key: ${{ runner.os }}-${{ runner.arch }}-scie-pants-v2
+          key: ${{ matrix.os }}-scie-pants-v4
```
jsirois, line 55: So, this is interesting. runner.os / runner.arch is the same for both our macOS runners - which is bad when used for a cache key (See "Configuration: macOS x64" towards the top for both):
+ https://github.com/organizations/pantsbuild/settings/actions/runners/818
+ https://github.com/organizations/pantsbuild/settings/actions/runners/819

@benjyw had you encountered this / known about it?

jsirois:
Ok, reviewers this is stable now. The macOS-X64 dup `runner.os`-`runner.arch` bogus cache key that routed macOS x86_64 and aarch64 runners to the same cache was very confusing but is now sussed (I don't understand why our mac arm self-hosted runner self-identifies this way, but at least I understand it does now).

.github/workflows/ci.yml
benjyw: Ugh. I'm guessing this is due to the fact that there used to be no native arm64 GHA runner binary, so we run the x86 one under rosetta.  But it looks like there is a native runner now, so I will switch to that.

benjyw (approved):

.github/workflows/ci.yml
jsirois: Ah, OK. That makes total sense then. Thanks.

PR #191, opened 2023-05-26
Build `scie-pants` using `science`.

jsirois:
Instead of using raw `scie-jump` and a hand-crafted JSON lift manifest,
use `science` and a TOML lift manifest to build `scie-pants`.

The `science` tool and its more friendly interface are the present and
future of end-user scie construction. We convert here and gain a few
things:
+ A manifest format we can comment.
+ Abstraction over platforms and PBS details.
+ Support for embedded build provenanace information for more
  comprehensive black box debugability.
+ A clean path to publishing both "thin" and "fat" binaries, solving
  #154.

PR #194, opened 2023-05-27
DRY up the package tool main.

jsirois:
This was noted tech debt in the conversion from raw scie-jump to
science PR here:
https://github.com/pantsbuild/scie-pants/pull/191#discussion_r1207115259

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,6 @@ jobs:
       - name: Build, Package & Integration Tests
         if: ${{ matrix.os == 'ubuntu-22.04' }}
         run: |
-          mkdir dist
           cargo run -p package -- --dest-dir dist/ tools
           docker run --rm \
             -v $PWD:/code \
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -81,13 +80,11 @@ jobs:
...
               '
-          cargo run -p package -- --dest-dir dist/ scie \
-            --scie-pants dist/scie-pants --tools-pex dist/tools.pex
           echo
           echo "Running under: $(uname -a)"
           echo
           PANTS_BOOTSTRAP_GITHUB_API_BEARER_TOKEN=${{ secrets.GITHUB_TOKEN }} \
             cargo run -p package -- test \
-              --tools-pex dist/tools.pex --scie-pants dist/scie-pants-linux-x86_64 \
+              --tools-pex dist/tools.pex --scie-pants dist/scie-pants \
```
jsirois, line 88: So this is the sort of thing a review eye might gloss over, but there is a small universe hidden here:
1. The default command of the scie-pants scie is to run the embedded scie-pants Rust binary: https://github.com/pantsbuild/scie-pants/blob/14ed78482a6e39d7f01a2d01e1b169c862c104ba/package/scie-pants.toml#L48-L51
2. On the LHS of this diff, the scie-pants scie was passed off as `--scie-pants` instead of the scie-pants Rust binary!

This mostly works because of 1 above except for env var manipulation by the enveloping scie. It was this double-manipulation of env vars (the outer scie-pants scie manipulates env vars and runs the default command which turns out not to be the scie-pants Rust binary but another scie-pants scie!) that led to a PANTS_BIN_NAME test failure.

huonw (approved):

stuhood (commented):
Thanks!

PR #219, opened 2023-07-03
Fix CI git describe error by busting the cache.

jsirois:
The error does not repro locally with cache or no; so try this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ${{ env.SCIE_PANTS_DEV_CACHE }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ${{ env.SCIE_PANTS_DEV_CACHE }}
-          key: ${{ matrix.os }}-scie-pants-v4
+          key: ${{ matrix.os }}-scie-pants-v5
```
jsirois, line 55: I have to hang my head in shame here - I have no clue why this is needed.

Failures looked like: https://github.com/pantsbuild/scie-pants/actions/runs/5440064753/jobs/9892625108#step:9:981

PR #225, opened 2023-07-12
Grab latest version from GitHub Release tag

thejcannon:
As discussed in the latest Pants meetup and [this proposal](https://github.com/pantsbuild/pants/discussions/19443), we're going to move off of PyPI for distribution, and instead leverage GitHub Releases.

Some reasons given for this pivot:
- People ought not to `pip install pantsbuild.pants`
- We ought not to continue eating up PyPI space, which we're already hitting the ceiling for
- We plan on having `scie-pants` use a PEX in the future, and PyPI wouldn't support that workflow
- We want to abstract away, as much is reasonable, that Pants is implemented in Python. And treat it like a black box, where possible.

Currently the latest version is expected to be set by hand. Although that's unlikely to be wrong, we're also changing to have the "latest" marker be [automated](https://github.com/pantsbuild/pants/pull/19444). Therefore we should be able to rely on it with confidence.

Tested and both before/after yielded the same string (with the GitHub URL performing slightly faster)

### DIFF
--- a/tools/src/scie_pants/pants_version.py
+++ b/tools/src/scie_pants/pants_version.py
@@ -149,11 +149,15 @@ def determine_latest_stable_version(
 ) -> tuple[Callable[[], None], ResolveInfo]:
     info(f"Fetching latest stable Pants version since none is configured")
 
-    latest_tag = ptex.fetch_json("https://github.com/pantsbuild/pants/releases/latest", Accept="application/json")["tag_name"]
+    latest_tag = ptex.fetch_json(
+        "https://github.com/pantsbuild/pants/releases/latest", Accept="application/json"
+    )["tag_name"]
     if not latest_tag.startswith("release_"):
-        fatal(f'Expected the GitHub Release tagged "latest" to have the "release_" prefix. Got "{latest_tag}"')
+        fatal(
+            f'Expected the GitHub Release tagged "latest" to have the "release_" prefix. Got "{latest_tag}"'
+        )
 
-    pants_version = latest_tag[len("release_"):]
+    pants_version = latest_tag[len("release_") :]
 
     def configure_version():
         backup = None
--- a/tools/src/scie_pants/pants_version.py
+++ b/tools/src/scie_pants/pants_version.py
@@ -149,12 +149,22 @@ def determine_latest_stable_version(
 ) -> tuple[Callable[[], None], ResolveInfo]:
     info(f"Fetching latest stable Pants version since none is configured")
 
-    latest_tag = ptex.fetch_json(
-        "https://github.com/pantsbuild/pants/releases/latest", Accept="application/json"
-    )["tag_name"]
+    try:
+        latest_tag = ptex.fetch_json(
+            "https://github.com/pantsbuild/pants/releases/latest", Accept="application/json"
+        )["tag_name"]
+    except Exception as e:
+        fatal(
+            "Couldn't get the latest release by fetching https://github.com/pantsbuild/pants/releases/latest.\n\n"
+            + "If this is unexpected (e.g. GitHub isn't down), please reach out on Slack: https://www.pantsbuild.org/docs/getting-help#slack\n\n"
+            + f"Exception:\n\n{e}"
+        )
+
     if not latest_tag.startswith("release_"):
         fatal(
-            f'Expected the GitHub Release tagged "latest" to have the "release_" prefix. Got "{latest_tag}"'
+            f'Expected the GitHub Release tagged "latest" to have the "release_" prefix. Got "{latest_tag}"\n\n'
+            + "Please reach out on Slack: https://www.pantsbuild.org/docs/getting-help#slack or file"
+            + " an issue on GitHub: https://github.com/pantsbuild/pants/issues/new/choose."
         )
 
     pants_version = latest_tag[len("release_") :]
--- a/tools/src/scie_pants/pants_version.py
+++ b/tools/src/scie_pants/pants_version.py
@@ -168,7 +168,6 @@ def determine_latest_stable_version(
             + " an issue on GitHub: https://github.com/pantsbuild/pants/issues/new/choose."
         )
 
-
     def configure_version():
         backup = None
         if pants_config.exists():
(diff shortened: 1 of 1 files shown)


tools/src/scie_pants/pants_version.py
```diff
@@ -148,9 +148,12 @@ def determine_latest_stable_version(
     ptex: Ptex, pants_config: Path, find_links_dir: Path, github_api_bearer_token: str | None = None
 ) -> tuple[Callable[[], None], ResolveInfo]:
     info(f"Fetching latest stable Pants version since none is configured")
-    pants_version = ptex.fetch_json("https://pypi.org/pypi/pantsbuild.pants/json")["info"][
-        "version"
-    ]
+
+    latest_tag = ptex.fetch_json("https://github.com/pantsbuild/pants/releases/latest", Accept="application/json")["tag_name"]
```
thejcannon, line 152: The good news is this is a normal public URL, and therefore (I assume) doesn't participate in the anonymous API rate limiting :tada:

jsirois:
I'll bow out. @kaos is a good reviewer for this I think.

tools/src/scie_pants/pants_version.py
```diff
@@ -148,9 +148,16 @@ def determine_latest_stable_version(
...
-    pants_version = ptex.fetch_json("https://pypi.org/pypi/pantsbuild.pants/json")["info"][
-        "version"
-    ]
+
+    latest_tag = ptex.fetch_json(
+        "https://github.com/pantsbuild/pants/releases/latest", Accept="application/json"
+    )["tag_name"]
+    if not latest_tag.startswith("release_"):
+        fatal(
+            f'Expected the GitHub Release tagged "latest" to have the "release_" prefix. Got "{latest_tag}"'
```
kaos, line 157: Messages like these are potentially user facing, so would prefer it included a little more context and/or suggested next course of action. (like, this is not you it's a pants bug, please report; that sort of thing.)

tools/src/scie_pants/pants_version.py
```diff
@@ -148,9 +148,16 @@ def determine_latest_stable_version(
...
+        )
+
+    pants_version = latest_tag[len("release_") :]
```
kaos, line 160: Like to avoid this kind of prefix checking when possible. How about:
```suggestion
    prefix, _, pants_version = latest_tag.partition("_")
    if prefix != "release" or not pants_version:
        fatal(
            f'Expected the GitHub Release tagged "latest" to have the "release_" prefix. Got "{latest_tag}"'
        )
```

tools/src/scie_pants/pants_version.py
```diff
@@ -148,9 +148,16 @@ def determine_latest_stable_version(
     ptex: Ptex, pants_config: Path, find_links_dir: Path, github_api_bearer_token: str | None = None
 ) -> tuple[Callable[[], None], ResolveInfo]:
     info(f"Fetching latest stable Pants version since none is configured")
-    pants_version = ptex.fetch_json("https://pypi.org/pypi/pantsbuild.pants/json")["info"][
-        "version"
-    ]
+
+    latest_tag = ptex.fetch_json(
+        "https://github.com/pantsbuild/pants/releases/latest", Accept="application/json"
+    )["tag_name"]
```
kaos, line 154: Come to think of it, could we avoid assuming the response will always include the `tag_name`, even if that is the case today.

The day that is not the case, the error is going to be _ugly_, and we can easily do a `.get("tag_name")` instead and check that `latest_tag != None` with a proper error message.

tools/src/scie_pants/pants_version.py
thejcannon: Idk. I think the prefix checking is a little repetitive, but is closer to the intent we're trying to translate into code.

Quite literally, what I want to represent (in English) is "is this tag doesn't start with 'release_' something is wrong, otherwise the version is the rest of the tag"

And this code _directly_ translates to that (which usually code doesn't get to directly translate from intent)

If we had Python 3.9, we could probably use `stripprefix`, but we don't.

thejcannon: Also your code assumes the tag will have 3 parts. But the whole point of checking is that it may not be what we expect. So that'll result in a TypeError (or whatever the unpack equivalent is)

tools/src/scie_pants/pants_version.py
thejcannon: Yes. It uses `json` to parse.

And I'd be _really_ surprised if GitHub changed their response schema like that. They seem to be very good REST citizens. But I'm also happy to tweak.

tools/src/scie_pants/pants_version.py
kaos: no, my code has no assumptions besides it being a string. `str.partition()` always returns a 3-tuple. If there are multiple instances of the partitioning charcter, the first will be the split point and the the rest preserved in the third element (the version, in this case).

My code reads, literally: "Split this string into two at the first occurrence of `_`, if the first part is "release" and we have a second part, that is the version, fail otherwise" (which basically is the same as yours using more words, but closer to what the code actually does.)

I like this more than what you have based on: fewer lines (while still being readable), not repeating the `"release_"` prefix, `len("release_")` is icky (but better than `8`.)

tools/src/scie_pants/pants_version.py
kaos: agree it's overly defensive, but also easy enough to do, and I prefer to protect against changes we have no control over.

tools/src/scie_pants/pants_version.py
thejcannon: I'm convinced. :+1:

tools/src/scie_pants/pants_version.py
```diff
@@ -148,9 +148,27 @@ def determine_latest_stable_version(
...
+
+    prefix, _, pants_version = latest_tag.partition("_")
+    if prefix != "release" or not pants_version:
+        fatal(
+            f'Expected the GitHub Release tagged "latest" to have the "release_" prefix. Got "{latest_tag}"\n\n'
+            + "Please reach out on Slack: https://www.pantsbuild.org/docs/getting-help#slack or file"
+            + " an issue on GitHub: https://github.com/pantsbuild/pants/issues/new/choose."
+        )
+
+    pants_version = latest_tag[len("release_") :]
```
thejcannon, line 171:
```suggestion
```

kaos (approved):
🚀

PR #226, opened 2023-07-12
Switch scie-pants to use Pants PEX, when supported

thejcannon:
This change makes it so `scie-pants` will create the Pants venv using the per-platform PEX uploaded to the release (for Pants releases). 

As a result, it bypasses the `--find-links` support that was otherwise required, so a few tweaks were made to allow that.

Additionally, we still install `debugpy` ad-hoc into the venv if required.

---

Using this repo, setting the version to `2.18.0.dev5`

```console
josh@cephandrius:~/work/scie-pants$ cargo run -p package -- scie
josh@cephandrius:~/work/scie-pants$ rm -rf /home/josh/.cache/nce/5b2404164fc1c3e3a34d18a5da428b73e51c82e6b5a7a020d563cb9bb1a9db20
josh@cephandrius:~/work/scie-pants$ dist/scie-pants-linux-x86_64 --version --no-verify-config
... (specifically I see it downloading the PEX)
New virtual environment successfully created at /home/josh/.cache/nce/0a5efe87c1f41645e4c37ec0402c76133d73a1246f35d59432dad968e1977d99/bindings/venvs/2.18.0.dev5.
2.18.0.dev5
```

### DIFF
--- a/tools/src/scie_pants/configure_pants.py
+++ b/tools/src/scie_pants/configure_pants.py
@@ -86,6 +86,7 @@ def main() -> NoReturn:
         resolve_info = determine_sha_version(
             ptex=ptex, sha=options.pants_sha, find_links_dir=find_links_dir
         )
+        assert resolve_info.sha_version is not None
         version = resolve_info.sha_version
     elif options.pants_version:
         resolve_info = determine_tag_version(
@@ -122,10 +123,11 @@ def main() -> NoReturn:
         finalizer()
 
     with open(env_file, "a") as fp:
-        print(f"FIND_LINKS={resolve_info.find_links}", file=fp)
+        if resolve_info.find_links:
+            print(f"FIND_LINKS={resolve_info.find_links}", file=fp)
+            print(f"PANTS_SHA_FIND_LINKS={resolve_info.pants_find_links_option(version)}", file=fp)
         if newly_created_build_root:
             print(f"PANTS_BUILDROOT_OVERRIDE={newly_created_build_root}", file=fp)
-        print(f"PANTS_SHA_FIND_LINKS={resolve_info.pants_find_links_option(version)}", file=fp)
         print(f"PANTS_VERSION={version}", file=fp)
         print(f"PYTHON={python}", file=fp)
 
--- a/tools/src/scie_pants/install_pants.py
+++ b/tools/src/scie_pants/install_pants.py
@@ -92,11 +92,12 @@ def install_pants_from_pex(
         except subprocess.CalledProcessError as e:
             fatal(
                 f"Wasn't able to fetch the Pants PEX at {pex_url}.\n\n"
-                + "Check to see if the URL is reachable (i.e. GitHub isn't down) and if the asset exists."
-                + " If the asset doesn't exist it may be that this platform isn't yet supported."
-                + " If that's the case, please reach out on Slack: https://www.pantsbuild.org/docs/getting-help#slack"
-                + " or file an issue on GitHub: https://github.com/pantsbuild/pants/issues/new/choose.\n\n"
-                + f"Exception:\n\n{e}"
+                "Check to see if the URL is reachable (i.e. GitHub isn't down) and if"
+                f" {pex_name} asset exists within the release: https://github.com/pantsbuild/pants/releases/release_{version}."
+                " If the asset doesn't exist it may be that this platform isn't yet supported."
+                " If that's the case, please reach out on Slack: https://www.pantsbuild.org/docs/getting-help#slack"
+                " or file an issue on GitHub: https://github.com/pantsbuild/pants/issues/new/choose.\n\n"
+                f"Exception:\n\n{e}"
             )
         subprocess.run(
             args=[
--- a/tools/src/scie_pants/install_pants.py
+++ b/tools/src/scie_pants/install_pants.py
@@ -8,6 +8,7 @@ import os
 import stat
 import subprocess
 import sys
+import tempfile
 from argparse import ArgumentParser
 from glob import glob
 from pathlib import Path
@@ -16,6 +17,8 @@ from typing import Iterable, NoReturn
 from packaging.version import Version
 
 from scie_pants.log import fatal, info, init_logging
+from scie_pants.pants_version import PANTS_PEX_GITHUB_RELEASE_VERSION
+from scie_pants.ptex import Ptex
 
 log = logging.getLogger(__name__)
 
@@ -41,7 +44,7 @@ def venv_pip_install(venv_dir: Path, *args: str, find_links: str | None) -> None
     )
 
 
-def install_pants(
+def install_pants_from_req(
     venv_dir: Path, prompt: str, pants_requirements: Iterable[str], find_links: str | None
 ) -> None:
     subprocess.run(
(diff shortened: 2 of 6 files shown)


thejcannon:
Marked as draft, since https://github.com/pantsbuild/scie-pants/pull/225 likely should be merged first, and https://github.com/pantsbuild/pants/pull/19450 definitely needs to be merged, with a release having happened.

tools/src/scie_pants/pants_version.py
```diff
@@ -18,19 +18,27 @@
...
 
-from scie_pants.log import info, warn
+from scie_pants.log import fatal, info, warn
 from scie_pants.ptex import Ptex
 
 log = logging.getLogger(__name__)
 
+# After this version, Pants is released as a per-platform PEX using GitHub Release assets.
+# See https://github.com/pantsbuild/pants/pull/19450
+PANTS_PEX_GITHUB_RELEASE_VERSION = Version("2.18.0.dev3")
```
thejcannon, line 28: This'll need to be adjusted to the actual right version for when the PEXs were released.
Additionally, the comment should reference the commit on `main`.

(We could technically backport this onto older releases and adjust this version downward. But that's a separate discussion)

thejcannon:
I tested this by modifying the URL to point to a release on a dummy release on a dummy repo. Happy to report it works as expected with/without debug.

kaos (approved):
lgtm

tools/src/scie_pants/install_pants.py
```diff
@@ -67,8 +67,58 @@ def pip_install(*args: str) -> None:
...
+            args=[
+                sys.executable,
+                pants_pex.name,
+                "venv",
+                "--prompt",
+                prompt,
+                "--compile",
+                "--pip",
+                "--collisions-ok",
+                # @TODO: Remove pex? --rm pex
```
thejcannon, line 111: @jsirois I couldn't figure out what the pros/cons of removing pex from the venv would be.

I assume that removing it would make a venv that is most like the existing venv today, so I lean towards removing it. But I'd love your expertise.

thejcannon: Nevermind, adding this causes an error (I suspect because the pex isn't in the venv) so ignore me.

tools/src/scie_pants/install_pants.py
```diff
@@ -67,8 +67,58 @@ def pip_install(*args: str) -> None:
...
+    """Installs Pants into the venv using the platform-specific pre-built PEX."""
+    uname = os.uname()
+    major, minor = sys.version_info[:2]
+    pex_name = f"pants.cp{major}{minor}-{uname.sysname.lower()}_{uname.machine.lower()}.pex"
+    with tempfile.NamedTemporaryFile(suffix=".pex") as pants_pex:
+        pex_url = (
+            f"https://github.com/pantsbuild/pants/releases/download/release_{version}/{pex_name}"
+        )
+        try:
+            ptex.fetch_to_fp(pex_url, pants_pex.file)
```
thejcannon, line 91: OK I think the last thing I'd love to see here is that the PEX is cached. If I have multiple Pants repos, but they use the same version, it'd be nice to have the PEX be cached.

But I'll call it a "nice-to-have" and just open an issue.

thejcannon:
Hmmm I think CI might be bitrot: https://pypi.org/project/pantsbuild.pants/2.16.0.dev5/

tools/src/scie_pants/ptex.py
```diff
@@ -45,5 +45,5 @@ def fetch_json(self, url: str, **headers: str) -> Any:
     def fetch_text(self, url: str, **headers: str) -> str:
         return self._fetch(url, stdout=subprocess.PIPE, **headers).stdout.decode()
 
-    def fetch_to_fp(self, url: str, fp: BinaryIO, **headers: str) -> None:
+    def fetch_to_fp(self, url: str, fp: IO, **headers: str) -> None:
```
huonw, line 48: It feels incorrect to be potentially using text-IO for this, i.e. `BinaryIO` seems like the right type. What prompted this change?

tools/src/scie_pants/install_pants.py
```diff
@@ -67,8 +67,59 @@ def pip_install(*args: str) -> None:
...
+        except subprocess.CalledProcessError as e:
+            fatal(
+                f"Wasn't able to fetch the Pants PEX at {pex_url}.\n\n"
+                + "Check to see if the URL is reachable (i.e. GitHub isn't down) and if the asset exists."
```
huonw, line 95: It'd be good to suggest how one can check the asset exists:

```suggestion
                + f"Check to see if the URL is reachable (i.e. GitHub isn't down) and if {pex_name} asset exists within the {version} release."
```

And potentially even linking to the actual release page for a human to check.

Another route for failure is that the Pants release doesn't exist (e.g. someone tries `PANTS_VERSION=2.1234.5678`)? Or is that detected earlier?

tools/src/scie_pants/install_pants.py
```diff
@@ -67,8 +67,59 @@ def pip_install(*args: str) -> None:
...
+                + " If the asset doesn't exist it may be that this platform isn't yet supported."
+                + " If that's the case, please reach out on Slack: https://www.pantsbuild.org/docs/getting-help#slack"
+                + " or file an issue on GitHub: https://github.com/pantsbuild/pants/issues/new/choose.\n\n"
+                + f"Exception:\n\n{e}"
```
huonw, line 99: FYI, the rest of this codebase seems to use implicit adjacent string concatenation, i.e.

```suggestion
                " or file an issue on GitHub: https://github.com/pantsbuild/pants/issues/new/choose.\n\n"
                f"Exception:\n\n{e}"
```

etc.

huonw (commented):
Nice work! Exciting to see this progressing.

Assuming this is the final bit of work along these lines for now, and that it's ready for release, you could add the version bump and appropriate changelog entry so that it's ready for tagging once merged: https://github.com/pantsbuild/scie-pants/blob/main/RELEASE.md

tools/src/scie_pants/install_pants.py
thejcannon: ergg :vomiting_face:

thejcannon:
I suspect we might want to wait for the last potential change, which gets us off pantsbuild.binaries.org for all v2 release builds.
Then I think the release is good to go.

jsirois:
@thejcannon I'll continue to beg off all pants related reviews. I'll just be working on Pex, a-scie/* and other projects going forward.

huonw (approved):
Exciting. Looks good in theory, although I haven't tried it locally.  It sounds like you have, so ✅ I'm assuming you checked that it still works for old versions as well as new versions?

Have you considered also adding a test to `tests.rs` to validate the `2.18.0.dev5` behaviour?

tools/src/scie_pants/configure_pants.py
```diff
@@ -122,10 +123,11 @@ def main() -> NoReturn:
         finalizer()
 
     with open(env_file, "a") as fp:
-        print(f"FIND_LINKS={resolve_info.find_links}", file=fp)
+        if resolve_info.find_links:
+            print(f"FIND_LINKS={resolve_info.find_links}", file=fp)
+            print(f"PANTS_SHA_FIND_LINKS={resolve_info.pants_find_links_option(version)}", file=fp)
```
jsirois, line 128: Since this binding is used unconditionally as the 1st arg to the `pants` command, it should probably use `--python-repos-find-links=[]` as the else value to avoid passing the empty string as argv1.

PR #246, opened 2023-08-23
Propagate scie-pants' version into the pants invocation

huonw:
This sets the `SCIE_PANTS_VERSION` env var when invoking pants, so that pants can read it and encourage users to upgrade if they need to.

For example, 2.17 will be the last release that supports the PyPI-based distribution model and Pants 2.18 will require scie-pants 0.9 or greater. So, in https://github.com/pantsbuild/pants/issues/19600, we're planning to have pants 2.17 encourage users to upgrade `scie-pants` if it's too old.

This PR preps for the release too, by bumping the version and adding a changelog entry. As part of this, I noticed the 0.9.0 release was accidentally configured as 0.9.1 in Cargo.toml, so updated the changelog to mention this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,20 @@
 # Release Notes
 
-## 0.9.0
+## 0.9.2
+
+This release propagates the version of the `scie-pants` into the invocation of pants, so that Pants
+itself can [detect if it requires (or will require) features from newer versions of
+`scie-pants`](https://github.com/pantsbuild/pants/issues/19600).
+
+## 0.9.0 / 0.9.1
 
 This release adds support for scie-pants to bootstrap Pants using a "per-platform" PEX uploaded as
 a GitHub release asset. Additionally, the new wheels.pantsbuild.prg/simple wheels index has been wired
 up to support Pants releases back to 2.0.0.dev0. Lastly, the selection of the latest version is now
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -1,12 +1,21 @@
...
+
+## 0.9.0 / 0.9.1
 
 This release adds support for scie-pants to bootstrap Pants using a "per-platform" PEX uploaded as
 a GitHub release asset. Additionally, the new wheels.pantsbuild.prg/simple wheels index has been wired
 up to support Pants releases back to 2.0.0.dev0. Lastly, the selection of the latest version is now
 deferred to the GitHub Release with the "latest" tag.
 
+NB. this version was published as 0.9.0, but `PANTS_BOOTSTRAP_VERSION=report pants` reports itself
+as 0.9.1.
```
thejcannon, line 17: Weird

thejcannon (approved):

CHANGES.md
huonw: I think there was just a small mistake in setting the version in Cargo..toml in https://github.com/pantsbuild/scie-pants/pull/239. I don't think it causes problems, just thought it'd be sensible to call out why 0.9.1 seems to be missed so people don't get confused.

thejcannon: I have no clue how I messed that up, lol

PR #250, opened 2023-08-25
Pass --python-repos-find-links always, and deprecate PANTS_SHA

### LINKED ISSUE #249: `pants tailor --check ::` fails with `Unknown flag --check on global scope` with new PEX distribution mechanism
**Describe the bug**
In 2.18.0.dev5 and beyond, `pants tailor --check ::` stops working when run under scie-pants 0.9.0 or greater.

This invocation is likely to be widespread, as it is suggested by [docs](https://www.pantsbuild.org/docs/using-pants-in-ci#approach-2-run-over-everything) and [examples](https://github.com/pantsbuild/example-python/blob/751db8be2ed502dba733d8bc1c340064bb9d1995/.github/workflows/pants.yaml#L52). It'd be _really_ good to not break (at least, not without deprecation warnings)

```
Unknown flag --check on global scope
Did you mean --check-only or --tailor-check?
Use `pants help` to get help.
13:19:23.55 [ERROR] Unknown flags --check on global scope


Use --print-stacktrace for more error details and/or -ldebug for more logs. 
See https://www.pantsbuild.org/v2.18/docs/troubleshooting for common issues.
Consider reaching out for help: https://www.pantsbuild.org/v2.18/docs/getting-help
```

Reproducer:

```shell
cd $(mktemp -d)

cat > pants.toml <<EOF
[GLOBAL]
pants_version = "2.18.0.dev5"

backend_packages = ["pants.backend.python"]
EOF

pants tailor --check ::
```

Changing the version to 2.18.0.dev4 passes (no output).

**Pants version**
2.18.0.dev5 to 2.18.0a0 at the time of writing

scie-pants 0.9.0, 0.9.3

**OS**
macOS and Linux

**Additional info**

- I discovered by a "public repo" test run, e.g. https://github.com/pantsbuild/pants/actions/runs/5957288454/job/16159713022
- Changes between those releases: https://github.com/pantsbuild/pants/compare/release_2.18.0.dev4...release_2.18.0.dev5 . ~Nothing leaps out at me. I haven't run a `git bisect` though.~ See first comment: it's the switch to PEX distributions

huonw:
Currently, the `scie-pants` launcher sometimes needs to pass an additional argument to the underlying pants invocation, to be able to support `PANTS_SHA` and its custom wheel location. Effectively turning an invocation like `pants goal --arg` into `pants --pants-repos-find-links=+["..."] goal --arg` when appropriate: https://github.com/pantsbuild/scie-pants/blob/0a204c5ac816ad24d74b55ecf7424397a07b3c2a/package/scie-pants.toml#L57-L60

However, when not using `PANTS_SHA`, the pants command becomes `pants "" goal --arg`, and the spurious empty string results in some commands failing, e.g. `pants "" tailor --check` thinks `--check` is a global argument, and this is ambiguous.

To fix this issue, this PR does two things:

1. avoid passing an empty string, via a no-op workaround: always set the `PANTS_SHA_FIND_LINKS` variable (to a no-op like `--python-repos-find-links=-[]`), so that scie-pants never passes an empty variable to the pants command, since that makes commands like `pants tailor --check ::` explode:
2. deprecates `PANTS_SHA`, so that the hack above doesn't feel so bad: we're on the way to removing the need for `PANTS_SHA_FIND_LINKS` entirely, thus sidestepping the limitation and the need for the workaround (2)

This could also be fixed via https://github.com/a-scie/jump/issues/130.

Fixes #249

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,20 @@
 # Release Notes
 
+## 0.10.0
+
+This release deprecates support for running against an arbitrary Pants commit using
+`PANTS_SHA=abc123... pants ...`. Pants no longer
+publishes the artifacts required for this for new commits, and so this is becoming less and less
+useful. To replace use of `PANTS_SHA`, do one of:
+
+- Use a released version of Pants.
+- Run pants from sources (for example: `PANTS_SOURCE=/path/to/pants-checkout pants ...`).
+- If these are not appropriate, [let us know what you're using it for](https://www.pantsbuild.org/docs/getting-help).
+
+
+This release also fixes scie-pants running commands like `pants tailor --check ...` with Pants
+2.18.0.dev5 and newer (releases that use the new "per-platform" PEX).
+
 ## 0.9.3
 
 This release propagates the version of the `scie-pants` into the invocation of pants, so that Pants
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -605,7 +605,7 @@ dependencies = [
 
 [[package]]
 name = "scie-pants"
-version = "0.9.3"
+version = "0.10.0"
 dependencies = [
  "anyhow",
  "dirs",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ members = [
 [package]
 name = "scie-pants"
 description = "Protects your Pants from the elements."
-version = "0.9.3"
+version = "0.10.0"
 edition = "2021"
 authors = [
     "John Sirois <john.sirois@gmail.com>",
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,7 +8,7 @@ use std::path::PathBuf;
 
 use anyhow::{anyhow, bail, Context, Result};
 use build_root::BuildRoot;
-use log::{info, trace};
+use log::{info, trace, warn};
 use logging_timer::{time, timer, Level};
 use uuid::Uuid;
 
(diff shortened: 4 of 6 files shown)


benjyw (approved):

huonw:
Just noting some manual testing:

I've confirmed the `execve` syscall in the same way as https://github.com/pantsbuild/scie-pants/issues/249#issuecomment-1692658229:

```
[
  "/root/.cache/nce/0de4b72a5e712632a9421fa33e3c3c0e072a2b7ed9babda918e01bd01164cdbc/bindings/venvs/2.18.0.dev5/bin/pants",
  "--python-repos-find-links=-[]",
  "--no-pantsd",
  "tailor",
  "--check",
  "::"
]
```

This doesn't seem to interfere with additional manual instances of that arg, e.g. `PANTS_PYTHON_REPOS_FIND_LINKS='+["abc"]' ./scie-pants-linux-aarch64 --python-repos-find-links='+["def"]' help-advanced python-repos` says:

```
  --python-repos-find-links="['<str>', '<str>', ...]"
  PANTS_PYTHON_REPOS_FIND_LINKS
  find_links
      default: []
      current value: [
          "abc",
          "def"
      ] (from env var PANTS_PYTHON_REPOS_FIND_LINKS, from command-line flag)
```

thejcannon:
Clever, but in the "we need to get unlocked" way 👍

thejcannon:
Also, does this mean we should bump the version number of `scie-pants` we check for, and cherry pick to 2.17.x?

😬

thejcannon:
> Also, does this mean we should bump the version number of `scie-pants` we check for, and cherry pick to 2.17.x?
> 
> 😬

https://github.com/pantsbuild/pants/pull/19670

PR #260, opened 2023-08-28
Consume versioned PEX releases

thejcannon:
See https://github.com/pantsbuild/pants/pull/19683 for the related Pantsbuild automation change.

This change makes it so we now consume PEX names with the expected Pants version in them.

This change isn't guarded by version, as I expect to backport this change to earlier releases.


thejcannon:
Looks like the tests will fail until the backporting happens.

kaos (approved):

PR #261, opened 2023-08-29
Allow firewalls for Pants PEXs

thejcannon:
This change extends the `PANTS_BOOTSTRAP_URLS` functionality to the fetched Pants PEX.

```console
josh@cephandrius:~/work/scie-pants$ PANTS_BOOTSTRAP_URLS=urls.json  PANTS_VERSION=2.22.0 dist/scie-pants-linux-x86_64
Bootstrapping Pants 2.22.0 using cpython 3.9.16
Installing pantsbuild.pants==2.22.0 into a virtual environment at /home/josh/.cache/nce/67dd6374fe21b02fa9e1be3a775826ef7fcac9791c6052276d2b3972a363d49f/bindings/venvs/2.22.0
Failed to fetch https://poop.com: [60] SSL peer certificate or SSH remote key was not OK (SSL certificate problem: unable to get local issuer certificate)
Wasn't able to fetch the Pants PEX at https://poop.com.
...
```


thejcannon:
FYI @kingjamm

package/scie-pants.toml
```diff
@@ -218,6 +218,8 @@ args = [
     "{scie.env.PANTS_DEBUG}",
     "--debugpy-requirement",
     "{scie.env.PANTS_DEBUGPY_VERSION}",
+    "--pants-pex-release-url-base",
+    "{scie.env.SCIE_PANTS_PANTS_PEX_RELEASE_URL_BASE}",
```
kaos, line 222: Does it have to be this value, or could we collapse the doubled `PANTS` in there.. ?

```suggestion
    "{scie.env.SCIE_PANTS_PEX_RELEASE_URL_BASE}",
```

kaos (approved):

package/scie-pants.toml
thejcannon: I think we get to decide. I see this as an env varname equivalent of something like `scie_pants::pants_pex::release_url_base`

kaos: Yea, I see the pattern it comes from. I'm just asking if we couldn't fold the double pants, much like you'd fold tripple consonants into just two when you write together two words. Seeing the namespaces are kind of loosely related to the name and not enforced.. 👀  I'm in favour of brevity/readability here..

thejcannon: Oh the other ambuguity is `scie-pants` does also download Pex from GitHub: https://github.com/pantsbuild/scie-pants/blob/891972ba3cb75f2bb7fe383927f3f25822e4c826/package/pbt.toml#L24

:grimacing: So the ambiguity would be is if this was "scie-pants PEX release url base" or "scie pants-PEX release url base"

thejcannon: (Not saying I love the double-Pants)

thejcannon: We could also try out other names. My brain isn't coming up with much...

kaos: Oh, right. Yea I think it may be a mistake to include PEX in the var name here (took me until now to realize exactly what it all meant.. granted my head is extra slow today but still).

Equally, scie isn't really interesting either. Why not attempt to get at what it is for only: `PANTS_BOOTSTRAP_BASE_URL` or some such..?

kaos: then comes the considerations of the namespaces that squats on (the BOOTSTRAP, in this case)

thejcannon: I'm worried about conflict with PANTS_ env vars from options

kaos: maybe `PANTS_INSTALL_BASE_URL` ?

kaos: > I'm worried about conflict with PANTS_ env vars from options

yep, that was my thought two seconds after posting too... but if we can get out of that?

kaos: `DOWNLOAD_PANTS_BASE_URL` 🤪

huonw: Oooh a bikeshed, let me join in!

I like the `PANTS_BOOTSTRAP_...` idea from @kaos : for better or for worse, scie-pants already uses that namespace, so seems sensible to continue using that (i.e. conflicts with `PANTS_...` env vars already already a thing).

These env vars already exist:

- `PANTS_BOOTSTRAP_GITHUB_API_BEARER_TOKEN`
- `PANTS_BOOTSTRAP_TOOLS`
- `PANTS_BOOTSTRAP_VERSION`

(There's one reference to `PANTS_BOOTSTRAP_URLS` too, but I don't know what reads it)

So, `PANTS_BOOTSTRAP_BASE_URL` or some variation sounds good to me!

huonw: Ah, actually: `PANTS_BOOTSTRAP_URLS` is referenced/described in https://github.com/pantsbuild/scie-pants?tab=readme-ov-file#features

So, maybe something could be riffed on there.

(Also, those docs should be updated in this PR, as appropriate.)

thejcannon: Ohh ok we're getting somewhere I think, that's really good info @huon

PANTS_BOOTSTRAP_PANTS_PEX_BASE_URL?

thejcannon: > So, maybe something could be riffed on there.

Hmm yeah let me try that :+1:

thejcannon: DONE! :rocket:

tools/src/scie_pants/install_pants.py
```diff
@@ -78,22 +79,30 @@ def install_pants_from_pex(
...
     find_links: str | None,
+    bootstrap_urls_path: str | None,
 ) -> None:
     """Installs Pants into the venv using the platform-specific pre-built PEX."""
     uname = os.uname()
-    major, minor = sys.version_info[:2]
-    pex_name = (
-        f"pants.{version}-cp{major}{minor}-{uname.sysname.lower()}_{uname.machine.lower()}.pex"
-    )
+    pex_name = f"pants.{version}-cp39-{uname.sysname.lower()}_{uname.machine.lower()}.pex"
```
thejcannon, line 86: Drive-by fix where we hardcode `cp39` because At some point Pants will run on 3.11, but this PEX can run on anything. SO hardcoding is safer.

kaos (approved):
Awesome 🚀

README.md
```diff
@@ -62,34 +62,39 @@ provides the following:
...
   }
   ```
-  To see the current mapping used by your version of `scie-pants` you can run:
+  To see the current mapping used by your version of `scie-pants` for Python Build Standalone you
+  can run:
   ```
   $ SCIE=inspect scie-pants | jq .ptex
   ```
   You'll need to run this once for each platform you use `scie-pants` on to gather all mappings
   you'll need; e.g.: once for Linux x86_64 and once for Mac ARM.
```
huonw, line 92: I think this suggestion won't reveal the pants pex names or the original/canonical URLs the user needs to download and supply. Is there some way we can hook that up or some alternative we can offer? Working it out manually seems like hard work...

README.md
```diff
@@ -62,34 +62,39 @@ provides the following:
...
 
   The keys in your re-mapping must match, but the URLs, of course will be different; presumably from
-  a private network server or file share. The full output of the inspect command can be used to 
-  examine the expected file size and hash of each of these distributions. Your re-directed URLs must
-  provide the same content; if the hashes of downloaded files do not match those recorded in
-  scie-pants, install will fail fast and let you know about the hash mismatch. Once Pants itself
-  starts shipping scies, those will also be able to redirected using the same file.
+  a private network server or file share. Your re-directed URLs must provide the same content as
+  the canonical URLs; if the hashes of downloaded files do not match those recorded in
+  scie-pants, install will fail fast and let you know about the hash mismatch.
```
huonw, line 97: Technically, the pants PEXes aren't covered by this, since we don't publish hashes and we couldn't hard code them into scie-pants anyway. Is it worth tweaking the text slightly?

tools/src/scie_pants/install_pants.py
```diff
@@ -78,22 +79,30 @@ def install_pants_from_pex(
...
+
     pex_url = f"https://github.com/pantsbuild/pants/releases/download/release_{version}/{pex_name}"
+    if bootstrap_urls_path:
+        bootstrap_urls = json.loads(Path(bootstrap_urls_path).read_text())
+        urls_info = bootstrap_urls["ptex"]
+        pex_url = urls_info.get(pex_name)
```
huonw, line 92: What's the behaviour/diagnostics if the JSON file is invalid? Eg invalid syntax, no `ptex` key, `ptex` key maps to non-object (eg `null`), the pex name maps to a non-string

huonw (commented):
Woohoo, nice!

README.md
thejcannon: To this point and the next, yeah I struggled to edit the wording to be inclusive of all the aspects of this new JSON file being partially for the `scie-pants` bootstrapping and partially for Pants PEX downloading.

So I'm open to suggestions, but this was my result.

---

Since the `SCIE=inspect` comes from the upstream `scie` tools, so I don't expect to find a way to change that. I'm expecting that if someone needs this, they'll hit some kind of error in their environment and be able to work backwards from there.

tools/src/scie_pants/install_pants.py
thejcannon: Everything but the last one, I assume is already handled by `scie` machinery itself. The very last one I assume is "garbage-in-garbage-out".

README.md
thejcannon: OK I spent a little time breaking it down father in this section

tools/src/scie_pants/install_pants.py
huonw: Hm, I think we can aspire to have better diagnostics in our user-friendly interactive tool than just "garbage-in-garbage-out". I tried the various cases like invalid JSON and so on, and they  just gave the raw `json` exception summary. There's a backtrace in `install.log`, but we can do better, e.g. you've already done great with the diagnostic for the key being missing.   I'd love to see that for other cases too. 😄 

As a relatively low effort improvement, we could do a whole-sale wrapping of any exception, plus checking for a string would be fine as a start:

```python
try:
    text = json.loads(Path(bootstrap_urls_path).read_text())
    urls_info = bootstrap_urls["ptex"]
    pex_url = urls_info.get(pex_name)
except Exception as e:
    raise ValueError(f"Failed to parse PANTS_BOOTSTRAP_URLS file '{bootstrap_urls_path}': it must be JSON like `{{"ptex": {{"{pex_name}": "$url", ...}}}}`: {e}") from e

if pex_url is None:
    ...

if not isinstance(pex_url, str):
    raise ValueError(
        f"Found invalid '{pex_name}' value {pex_url!r} in PANTS_BOOTSTRAP_URLS file '{bootstrap_urls_path}': it must be a URL as a string"
    )
```

One possibility would be pulling in a library like Pydantic or attrs or similar to not have to do all this manual validation/parsing, but that's probably not worth it if it's just this one place.

tools/src/scie_pants/install_pants.py
```diff
@@ -78,22 +79,30 @@ def install_pants_from_pex(
...
+        if pex_url is None:
+            raise ValueError(
+                f"Couldn't find '{pex_name}' in '{bootstrap_urls_path}' under the 'ptex' key."
```
huonw, line 95: I personally find it helpful for tools to explain where they got values like this from, because it might be a random filename that's hard to grep for (as in, one might be confused about why scie-pants is looking in a particular file, having forgotten what's set in the environment). Hence:

```suggestion
                f"Couldn't find '{pex_name}' in PANTS_BOOTSTRAP_URLS file '{bootstrap_urls_path}' under the 'ptex' key."
```

huonw (approved):
Looks good, other than maybe improving the diagnostics.

What do you think about adding a test for this too?

tools/src/scie_pants/install_pants.py
thejcannon: When you tested invalid JSON, did you test it end-to-end? I expected the scie bits to have already errored by the time we get here on things like invalid JSON

thejcannon: Either way, I'll bang on this some more

huonw: Yeah, you're right, not fully-end-to-end, the scie/ptex stuff was already installed, so maybe not a fully realistic scenario.

thejcannon: OK I tested, and anything but the value being the wrong type is already handled by the `scie` plumbing through generic errors.

PR #269, opened 2023-09-07
Add some cruft files to gitignore.

benjyw:


thejcannon (approved):
Dunno what these are, but sure

benjyw:
`.DS_Store` is macos Finder metadata for the directory.

A tilde suffix is a vim or emacs backup file.

thejcannon:
I see 

I'm a purist when it comes to gitignore files. They should only ignore things that everyone will run into when interacting with the repo.

For everything else, i prefer user-gitignore configs. Otherwise these things tend to bloat and occasionally you start ignoring things you realize you want to check in. Then you're in a catch 22.

But I also know I'm probably in the minority in my opinion

benjyw:
Yeah, I disagree with that approach. IMO the job of .gitignore is to prevent crap ending up in the git history by mistake. And delegating that to every separate potential purveyor of crap seems like an accident waiting to happen.

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     runs-on: ${{ matrix.os }}
```
thejcannon, line 27: Did you mean to commit this?

thejcannon (dismissed):

.github/workflows/ci.yml
benjyw: https://github.com/pantsbuild/scie-pants/pull/269/commits/62cc6b23dd3d08dadfd7a767c3f1d84a53523a24

benjyw: So, yes

benjyw: I am debugging the CI failure

benjyw:
The CI failures are this: https://github.com/pantsbuild/scie-pants/issues/270

.github/workflows/ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ${{ env.SCIE_PANTS_DEV_CACHE }}
-          key: ${{ matrix.os }}-scie-pants-v5
+          key: ${{ matrix.os }}-scie-pants-v6
```
benjyw, line 55: This is #271 , which I will merge before this one.

benjyw:
@thejcannon PTAL now that the debugging changes have been reverted

thejcannon (approved):

PR #273, opened 2023-09-11
Upgrade Pants's PEX to 2.1.137.

benjyw:
To match the one embedded in pbt.

### DIFF
--- a/pants.toml
+++ b/pants.toml
@@ -25,6 +25,15 @@ resolves_generate_lockfiles = false
 [python.resolves]
 python-default = "tools/lock.json"
 
+[pex-cli]
+version = "v2.1.137"
+known_versions = [
+    "v2.1.137|macos_arm64 |faad51a6a108fba9d40b2a10e82a2646fccbaf8c3d9be47818f4bffae02d94b8|4098329",
+    "v2.1.137|macos_x86_64|faad51a6a108fba9d40b2a10e82a2646fccbaf8c3d9be47818f4bffae02d94b8|4098329",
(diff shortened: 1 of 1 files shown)


benjyw:
Before this change, `cargo run -p package -- test` failed with 

```
>> Verifying the tools.pex built by the package crate matches the tools.pex built by Pants
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `"8ee6822086594dd83394f921c8bac0a25a8d70fa530522d59064062f3ba2b63f"`,
 right: `"e7bf24a357bd0d959614df1d2516d088e8e44497a354c615d415104d6f830d14"`', package/src/test.rs:288:9
```

With this change, `cargo run -p package -- test` passes.

huonw (approved):

huonw:
Do you have ideas about why CI didn't catch this?

(I guess this may've been caused by https://github.com/pantsbuild/scie-pants/pull/259 ? although that was pinning PEX to 2.1.126, not .137 🤔 )

benjyw:
I spoke too soon, https://github.com/pantsbuild/scie-pants/pull/274 is also required to get identical tools.pex.

thejcannon:
If anything, I think we should have pbts version track the pants one... it'd be less to maintain

thejcannon:
Also how does this relate to my original post in #development?
https://pantsbuild.slack.com/archives/C0D7TNJHL/p1694101689659639?thread_ts=1694101689.659639&cid=C0D7TNJHL


The metadata in pbt..toml doesn't use 137, it uses 134

PR #274, opened 2023-09-11
Change Pants ICs to include 3.8.

benjyw:
Older pants versions require 3.8.

### DIFF
--- a/pants.toml
+++ b/pants.toml
@@ -17,7 +17,7 @@ config = "tools/pyproject.toml"
 
 [python]
 enable_resolves = true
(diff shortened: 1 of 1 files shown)


benjyw:
Now `cargo run -p package -- test` passes for real.

huonw (approved):

thejcannon (commented):
We should add a note that these ICs have to match the one in Rust land (tools_pex.rs) and that the lower bounds needs to be the lowest version of Python that scie-pants might provide

PR #282, opened 2023-09-19
avoid printing Pants Python version to the console

cburroughs:
This frequently causes onboarding confusion as the distinction between "the version of python pants uses" and "the version of python pants uses for my code uses" it not yet clear.  (And maybe should remain an implementation detail outside of plugin authors anyway.)

### DIFF
--- a/tools/src/scie_pants/install_pants.py
+++ b/tools/src/scie_pants/install_pants.py
@@ -176,7 +176,8 @@ def main() -> NoReturn:
 
     version = options.pants_version
     python_version = ".".join(map(str, sys.version_info[:3]))
-    debug(f"Bootstrapping Pants {version} using {sys.implementation.name} {python_version}")
+    info(f"Bootstrapping Pants {version}")
+    debug(f"Pants itself is using: {sys.implementation.name} {python_version}")
 
     pants_requirements = [f"pantsbuild.pants=={version}"]
     extra_requirements = []
--- a/tools/src/scie_pants/install_pants.py
+++ b/tools/src/scie_pants/install_pants.py
@@ -17,7 +17,7 @@ from typing import Iterable, NoReturn
 
 from packaging.version import Version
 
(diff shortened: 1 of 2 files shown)


tools/src/scie_pants/install_pants.py
```diff
@@ -176,7 +176,7 @@ def main() -> NoReturn:
 
     version = options.pants_version
     python_version = ".".join(map(str, sys.version_info[:3]))
-    info(f"Bootstrapping Pants {version} using {sys.implementation.name} {python_version}")
+    debug(f"Bootstrapping Pants {version} using {sys.implementation.name} {python_version}")
```
kaos, line 179: I think it would be nice to keep the first half, that is which version of pants is being bootstrapped. In case of lingering env vars or weird config files this could reveal surprising values.. or just serve as a reminder which version is actually in use.

kaos (approved):

tools/src/scie_pants/install_pants.py
cburroughs: image

Log file:
```
2023-09-19 16:20:58,873 INFO] root: Bootstrapping Pants 2.17.0
2023-09-19 16:20:58,873 DEBUG] root: Pants itself is using: cpython 3.9.16
2023-09-19 16:20:58,873 INFO] root: Installing pantsbuild.pants==2.17.0 into a virtual environment at /home/ecsb/.cache/nce/8a9caced2db752386b323ad28c99e9953dda86f7f75af8d8bbee62d4f3be11a7/bindings/venvs/2.17.0
```

kaos (approved):
Great 🚀

tools/src/scie_pants/install_pants.py
kaos: ah, true. didn't remember the Installing log line there... so would've been ok either way.

huonw (approved):
Looks great, thanks!

PR #284, opened 2023-09-20
Update jump & science to improve error message for bad SCIE_BOOT

huonw:
This updates to jump to 0.13.0, which improves the error message for a bad `SCIE_BOOT` value. For example `SCIE_BOOT='update pants' pants` from https://pantsbuild.slack.com/archives/C046T6T9U/p1695047533069949 now shows:

```
Error: `SCIE_BOOT=update pants` was found in the environment but "update pants" does not correspond to any scie-pants commands.

Isolates your Pants from the elements.

<default> (when SCIE_BOOT is not set in the environment)  Detects the current Pants installation and launches it.
bootstrap-tools                                           Introspection tools for the Pants bootstrap process.
update                                                    Update scie-pants.

You can select a boot command by setting the SCIE_BOOT environment variable.
```

This also requires updating science to 0.2.1, for the `descriptions` to be propagated appropriately.

### DIFF
--- a/package/pbt.toml
+++ b/package/pbt.toml
@@ -8,7 +8,7 @@ Python Build Tool: A BusyBox that provides `python`, `pip`, `pex`, `pex3` and `p
 version = "0.7.0"
 
 [lift.scie-jump]
-version = "0.11.1"
+version = "0.13.0"
 
 [[lift.interpreters]]
 id = "cpython"
--- a/package/pbt.toml
+++ b/package/pbt.toml
@@ -7,7 +7,7 @@ Python Build Tool: A BusyBox that provides `python`, `pip`, `pex`, `pex3` and `p
 [lift.ptex]
 version = "0.7.0"
 
-[lift.scie-jump]
+[lift.scie_jump]
 version = "0.13.0"
 
 [[lift.interpreters]]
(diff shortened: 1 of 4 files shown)


huonw:
Hm, the test in CI is failing: it now seems to include the descriptions correctly, but not the text about a bad `SCIE_BOOT` 🤔 Will have to look later.

package/pbt.toml
```diff
@@ -8,7 +8,7 @@ Python Build Tool: A BusyBox that provides `python`, `pip`, `pex`, `pex3` and `p
 version = "0.7.0"
 
 [lift.scie-jump]
```
jsirois, line 10: The issue is I wrote this badly from the get go. Should be `[lift.scie_jump]`. See the reference: https://science.scie.app/_/manifest/lift.html#scie-jump

huonw: Ah, thanks, that short cuts the investigation I hadn't got to! 🎉 

```suggestion
[lift.scie_jump]
```

package/scie-pants.toml
```diff
@@ -12,7 +12,7 @@ version = "0.7.0"
 lazy_argv1 = "{scie.env.PANTS_BOOTSTRAP_URLS={scie.lift}}"
 
 [lift.scie-jump]
```
huonw, line 14:
```suggestion
[lift.scie_jump]
```

kaos (approved):

package/pbt.toml
jsirois: To fix this sort of thing going forward: https://github.com/a-scie/lift/issues/46

PR #288, opened 2023-09-25
Prepare 0.10.2

huonw:
Includes #282 and #284, plus a bunch of Rust package upgrades.

PR #290, opened 2023-09-26
Require Cargo.lock matches Cargo.toml in CI

huonw:
This blocks the mistake I made in #288 / #289, where I updated `Cargo.toml` and forgot to run a command to update `Cargo.lock`.


huonw:
Validating that this does what we expect:

- the first run was before #289 merged and failed with `error: the lock file /home/runner/work/scie-pants/scie-pants/Cargo.lock needs to be updated but --locked was passed to prevent this` ✅ 
- the second run was after #289 merged, and passed ✅

PR #293, opened 2023-09-26
Obey `PANTS_BOOTSTRAP_URLS` properly, update to science 0.2.2

### LINKED ISSUE #243: PANTS_BOOTSTRAP_URLS is not honored by scie-pants
When the following file is created the scie-pants command does not honor the contents:

    {
      "ptex": {
        "cpython-3.8.16+20230507-x86_64-unknown-linux-gnu-install_only.tar.gz": "file:///home/build/ptex/cpython-3.8.16+20230507-x86_64-unknown-linux-gnu-install_only.tar.gz",
        "cpython-3.9.16+20230507-x86_64-unknown-linux-gnu-install_only.tar.gz": "file:///home/build/ptex/cpython-3.9.16+20230507-x86_64-unknown-linux-gnu-install_only.tar.gz"
      }
    }

Where the keys were derived from the `SCIE=inspect pants | jq .ptex` command. The values for the keys point to a local file system for the ptex runtimes.

After creation of this file running the following command yields the original values

`PANTS_BOOTSTRAP_URLS=/home/build/ptex-mapping.json SCIE=inspect pants`

    {
      "cpython-3.8.16+20230507-x86_64-unknown-linux-gnu-install_only.tar.gz": "https://github.com/indygreg/python-build-standalone/releases/download/20230507/cpython-3.8.16%2B20230507-x86_64-unknown-linux-gnu-install_only.tar.gz",
      "cpython-3.9.16+20230507-x86_64-unknown-linux-gnu-install_only.tar.gz": "https://github.com/indygreg/python-build-standalone/releases/download/20230507/cpython-3.9.16%2B20230507-x86_64-unknown-linux-gnu-install_only.tar.gz"
    }

The pants command in use was downloaded using the following:

    curl --proto '=https' --tlsv1.2 -fsSL https://static.pantsbuild.org/setup/get-pants.sh | bash

This was done as of August 15 2023

huonw:
This updates to https://github.com/a-scie/lift/releases/tag/v0.2.2, which has diagnostics to detect invalid/ignored entries in the configuration files. This detects two:

- the `scie-pants.bin` file wasn't being marked as executable properly... which suggests it doesn't need to be
- the `PANTS_BOOTSTRAP_URLS` env var wasn't being passed through to `ptex` properly, and thus redirecting the downloads wasn't working as intended.

Fixes #243

### DIFF
--- a/package/scie-pants.toml
+++ b/package/scie-pants.toml
@@ -9,7 +9,7 @@ repo = "https://github.com/pantsbuild/scie-pants"
 [lift.ptex]
 id = "ptex"
 version = "0.7.0"
-lazy_argv1 = "{scie.env.PANTS_BOOTSTRAP_URLS={scie.lift}}"
+argv1 = "{scie.env.PANTS_BOOTSTRAP_URLS={scie.lift}}"
 
 [lift.scie_jump]
 version = "0.13.0"
@@ -40,7 +40,7 @@ members = [
 # N.B.: We name the scie-pants binary scie-pants.bin since the scie itself is named scie-pants
 # which would conflict when packaging.
 name = "scie-pants.bin"
-executable = true
+is_executable = true
 
 [[lift.files]]
 name = "tools.pex"
--- a/package/scie-pants.toml
+++ b/package/scie-pants.toml
@@ -40,7 +40,6 @@ members = [
 # N.B.: We name the scie-pants binary scie-pants.bin since the scie itself is named scie-pants
 # which would conflict when packaging.
 name = "scie-pants.bin"
-is_executable = true
 
 [[lift.files]]
 name = "tools.pex"
--- a/package/src/main.rs
+++ b/package/src/main.rs
@@ -29,7 +29,7 @@ use crate::utils::fs::{base_name, canonicalize, copy, ensure_directory};
 
 const BINARY: &str = "scie-pants";
 
-const SCIENCE_TAG: &str = "v0.2.1";
+const SCIENCE_TAG: &str = "v0.2.2";
 
 #[derive(Clone)]
 struct SpecifiedPath(PathBuf);


package/scie-pants.toml
```diff
@@ -40,7 +40,7 @@ members = [
 # N.B.: We name the scie-pants binary scie-pants.bin since the scie itself is named scie-pants
 # which would conflict when packaging.
 name = "scie-pants.bin"
-executable = true
```
huonw, line 43: The error for this one referred to `files[1].executable` which appears to be 1-indexed (AFAICT, this is the first entry in the `files` list); is that expected?

huonw:
(CI is blocked because of #292.)

package/scie-pants.toml
jsirois: It's expected, but perhaps a bad choice. It carried over from my Pex lockfile parsing diagnostics. There I saw a large audience was comprised of ML folks and I've been continually surprised by how much ML they know combined with how little programming language familiarity they have. I should probably just add a parenthetical to diagnostic messages that array indexes are X-based.

jsirois (approved):

package/scie-pants.toml
```diff
@@ -40,7 +40,7 @@ members = [
...
+is_executable = true
```
jsirois, line 43: This one is apparently un-needed by implication.

package/scie-pants.toml
```diff
@@ -9,7 +9,7 @@ repo = "https://github.com/pantsbuild/scie-pants"
 [lift.ptex]
 id = "ptex"
 version = "0.7.0"
-lazy_argv1 = "{scie.env.PANTS_BOOTSTRAP_URLS={scie.lift}}"
+argv1 = "{scie.env.PANTS_BOOTSTRAP_URLS={scie.lift}}"
```
jsirois, line 12: This one was a real bug. You're supposed to be able to point at a private repo with the env var and this would have made that not work.

jsirois: There was additional confusion here re `SCIE=inspect`, but this may have been a bug sighting: https://pantsbuild.slack.com/archives/C046T6T9U/p1692201968356429?thread_ts=1692132192.942339&cid=C046T6T9U

jsirois: And, in fact, see the update here: https://github.com/pantsbuild/scie-pants/issues/243

huonw: Ah, good catch. I've confirmed that this indeed fixes the behaviour there.

PR #319, opened 2023-11-03
Update to scie jump 0.13.2, for errors on invalid .env

### LINKED ISSUE #307: .env loading silently fails under common scenarios
With the following test:

```python
## example_test.py

import os

def test_potato():
  assert os.getenv("POTATO") == "potato"

def test_cabbage():
  assert os.getenv("CABBAGE") == "cabbage"
```

```python
## BUILD
python_tests(extra_env_vars=["POTATO"])
```

This `.env` file works when running `pants test ::`
```
POTATO=potato
CABBAGE=cabbage
```

So does:

```
POTATO=potato
CABBAGE=cabbage
PYTHONPATH="/foo/bar:$PYTHONPATH"
```

But the following does not, and does not warn you there was a problem loading the env file.
```
CABBAGE=cabbage
PYTHONPATH="/foo/bar:$PYTHONPATH"
POTATO=potato
```

This is particularly troublesome, as adding `PYTHONPATH="...:$PYTHONPATH"` to your `.env` file is recommended to get vscode integration working in the documentation: https://www.pantsbuild.org/docs/setting-up-an-ide#first-party-sources

This seems to be the case for both Pants 2.16 and 2.17 .

huonw:
This fixes #307 by ensuring that an invalid `.env` file is flagged loudly, rather than silently ignored. This is done by updating scie jump to pick up https://github.com/a-scie/jump/issues/163 / https://github.com/a-scie/jump/pull/164 in https://github.com/a-scie/jump/releases/tag/v0.13.2 .

This is technically a breaking change, as someone may be using scie-pants fine, despite having an `.env` file lying around that's not understood by the underlying `dotenvy`. Thus, this is a bump to 0.11, rather than a patch to 0.10.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## 0.11.0
 
-This releases ensures that `scie-pants` emits an error if there's an `.env` file cannot be
+This release ensures that `scie-pants` emits an error if there's an `.env` file that cannot be
 understood, rather than silently ignoring the invalid contents (and everything after). In
 particular, `"`s are not supported, a line like `A="B C"` should be `A='B C'`.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Release Notes
 
+## 0.11.0
+
+This releases ensures that `scie-pants` emits an error if there's an `.env` file cannot be
+understood, rather than silently ignoring the invalid contents (and everything after). In
+particular, `"`s are not supported, a line like `A="B C"` should be `A='B C'`.
+
 ## 0.10.4
 
 This release fixes a regression when using `PANTS_DEBUG`. The bundled Python interpreter has also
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Release Notes
 
+## 0.11.0
+
+This releases ensures that `scie-pants` emits an error if there's an `.env` file cannot be
```
kaos, line 5:
```suggestion
This release ensures that `scie-pants` emits an error if there's an `.env` file that cannot be
```

kaos (approved):
lovely

huonw:
Note to self: I've reconsidered and think we shouldn't merge this as is, and instead make it less breaking before merging. Potentially both of:

- have this be a warning first, before being an error
- wait for support for at least `X="double quotes"`, since that's probably fairly common

package/src/test.rs
```diff
@@ -517,6 +518,26 @@ fn test_dot_env_loading(scie_pants_scie: &Path, clone_root: &TempDir) {
     .unwrap();
 }
 
+fn test_dot_env_error(scie_pants_scie: &Path) {
```
jsirois, line 521: I have no idea what to make of exchanges like this: https://github.com/pantsbuild/scie-pants/pull/339#issuecomment-1868017847

That PR got submitted trampling this test. The test seemed like a great idea to me.

huonw:
Replaced by #339 and now #389

PR #338, opened 2023-12-19
Add validation for pants_version format

### LINKED ISSUE #337: Clarification Needed on Error Message for Incorrect `pants_version` Format
When setting up Pants with a `pants.toml` file that specifies the version in `pants.toml` as pants_version = "2.17", 
```text
 [GLOBAL]
 pants_version = "2.17"
 ```
the following error message is generated:

```bash
$ pants tailor ::
Failed to fetch https://binaries.pantsbuild.org/tags/pantsbuild.pants/release_2.17: [22] HTTP response code said error (The requested URL returned error: 404)
Install failed: list indices must be integers or slices, not str
More information can be found in the log at: /Users/xxx/Library/Caches/nce/cf6c1988edaa5d40bbf385ae79a75e6d2abba937217d1b672c71eaf0ea495bcd/bindings/logs/configure.log

Error: Failed to establish atomic directory /Users/xxx/Library/Caches/nce/cf6c1988edaa5d40bbf385ae79a75e6d2abba937217d1b672c71eaf0ea495bcd/locks/configure-e2013035cf3930334d2e37fdaa72d908ce286bd0690a7afa8f80b214f3084c3d. Population of work directory failed: Boot binding command failed: exit status: 1
```

Issue:
The error message does not clearly indicate that the version number format is incorrect or suggest a resolution.

Solution:
Specifying the version number in `x.y.z` format, such as `2.17.0`, resolves the issue. A more informative error message, guiding the user to use the correct version format, would be helpful.

Seaweed0112:
Implemented a check to ensure the 'pants_version' follows the 'major.minor.micro' format. This update raises a clear error message if the version string does not meet the expected format, enhancing user guidance and preventing configuration errors.

Fix #337

### DIFF
--- a/tools/src/scie_pants/install_pants.py
+++ b/tools/src/scie_pants/install_pants.py
@@ -105,8 +105,15 @@ def install_pants_from_pex(
         try:
             ptex.fetch_to_fp(pex_url, pants_pex.file)
         except subprocess.CalledProcessError as e:
+            #  if there's only one dot in version, specifically suggest adding the `.patch`)
+            suggestion = (
+                "Pants version format not recognized. Please add `.<patch_version>` to the end of the version. For example: `2.18` -> `2.18.0`"
+                if version.base_version.count(".") < 2
+                else ""
+            )
             fatal(
                 f"Wasn't able to fetch the Pants PEX at {pex_url}.\n\n"
+                f"{suggestion}\n\n"
                 "Check to see if the URL is reachable (i.e. GitHub isn't down) and if"
                 f" {pex_name} asset exists within the release."
                 " If the asset doesn't exist it may be that this platform isn't yet supported."
--- a/tools/src/scie_pants/install_pants.py
+++ b/tools/src/scie_pants/install_pants.py
@@ -107,13 +107,12 @@ def install_pants_from_pex(
         except subprocess.CalledProcessError as e:
             #  if there's only one dot in version, specifically suggest adding the `.patch`)
             suggestion = (
-                "Pants version format not recognized. Please add `.<patch_version>` to the end of the version. For example: `2.18` -> `2.18.0`"
+                "Pants version format not recognized. Please add `.<patch_version>` to the end of the version. For example: `2.18` -> `2.18.0`.\n\n"
                 if version.base_version.count(".") < 2
                 else ""
             )
             fatal(
-                f"Wasn't able to fetch the Pants PEX at {pex_url}.\n\n"
-                f"{suggestion}\n\n"
+                f"Wasn't able to fetch the Pants PEX at {pex_url}.\n\n{suggestion}"
                 "Check to see if the URL is reachable (i.e. GitHub isn't down) and if"
                 f" {pex_name} asset exists within the release."
                 " If the asset doesn't exist it may be that this platform isn't yet supported."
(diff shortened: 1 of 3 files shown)


Seaweed0112:
Fix #337

tools/src/scie_pants/pants_version.py
```diff
@@ -99,6 +99,11 @@ def determine_find_links(
 def determine_tag_version(
     ptex: Ptex, pants_version: str, find_links_dir: Path, github_api_bearer_token: str | None = None
 ) -> ResolveInfo:
+    # `pants_version` should be in the format `major.minor.micro`, e.g., `2.17.0`.
+    # Raise an error if it does not follow this format.
+    if not pants_version.count(".") == 2:
```
huonw, line 104: FYI, a dev release has a format with more `.`s, so we'll need to refine this somewhat to support them. For instance: https://github.com/pantsbuild/pants/releases/tag/release_2.20.0.dev3

huonw (commented):
Thanks very much for taking the time to contribute!

As you noted on slack #156 is quite similar. I wonder if we could take more of a "post-hoc" explanation approach here, similar to that one. As in, 'trust' the version more, until we get to a point where a network request definitely fails, and then offer explanations of what might've gone wrong.

So, I think this means:

- for versions 2.17 and early, merging #156 and updating it (potentially doing things like: if there's only one dot, specifically suggest adding the `.patch`)
- for versions 2.18 and later, adjusting the following code to detect the missing patch too. https://github.com/pantsbuild/scie-pants/blob/76911604586dab64cc226c5891504be607b506da/tools/src/scie_pants/install_pants.py#L108-L115

It would also be good to add tests, similar to #156. 

(I'm a bit busy so I can't provide lots of info promptly, but please ask questions!)

tools/src/scie_pants/pants_version.py
Seaweed0112: I just improved the error message and added a test. Please take a look! Thank you.

tools/src/scie_pants/install_pants.py
```diff
@@ -105,8 +105,15 @@ def install_pants_from_pex(
         try:
             ptex.fetch_to_fp(pex_url, pants_pex.file)
         except subprocess.CalledProcessError as e:
+            #  if there's only one dot in version, specifically suggest adding the `.patch`)
+            suggestion = (
+                "Pants version format not recognized. Please add `.<patch_version>` to the end of the version. For example: `2.18` -> `2.18.0`"
```
huonw, line 110: I'd suggest we make two minor adjustments:

- have this be a full sentence (i.e. `.` on the end)
- ensure that if there's no suggestion, we don't end up with 3 blank lines, by including the `\n\n` here

```suggestion
                "Pants version format not recognized. Please add `.<patch_version>` to the end of the version. For example: `2.18` -> `2.18.0`.\n\n"
```

Then change `f"{suggestion}\n\n"` to be just `f"{suggestion}"` or maybe include `{suggestion}` at the end of the first line.

Seaweed0112: Thank so much! I just pushed another commit. Please check!

huonw (approved):
Thanks for this!

PR #351, opened 2024-01-22
Prep support for running Pants with Python 3.10 or 3.11.

kaos:
~Look at the Python tag in the Pants PEX filename~ Cycle through all supported Python versions and check the PEX download URL using HEAD for each one to infer the Python version to use when installing Pants, when using a version of Pants that we don't already know which version of Python to use.

~Drop support for `PANTS_SHA`.~ Done in #376.

~Silence the install step, capturing the output to a `pants-install.log` file in the science cache.~ Done in #375.


kaos (commented):
@thejcannon if you're able, would love to get your eyes on my assumptions about our released pex files 😬 in case I'm wrong about anything here..

tools/src/scie_pants/install_pants.py
```diff
@@ -105,43 +73,48 @@ def install_pants_from_pex(
...
             ptex.fetch_to_fp(pex_url, pants_pex.file)
         except subprocess.CalledProcessError as e:
-            #  if there's only one dot in version, specifically suggest adding the `.patch`)
-            suggestion = (
-                "Pants version format not recognized. Please add `.<patch_version>` to the end of the version. For example: `2.18` -> `2.18.0`.\n\n"
-                if version.base_version.count(".") < 2
-                else ""
-            )
             fatal(
-                f"Wasn't able to fetch the Pants PEX at {pex_url}.\n\n{suggestion}"
```
kaos, line 115: The suggestion has moved to the configuration binding, as an unknown version will fail at that point and never get here.

tools/src/scie_pants/install_pants.py
```diff
@@ -105,43 +73,48 @@ def install_pants_from_pex(
...
+                    "--compile",
+                    "--pip",
+                    "--collisions-ok",
+                    "--no-emit-warnings",  # Silence `PEXWarning: You asked for --pip ...`
+                    "--disable-cache",
+                    str(venv_dir),
+                ],
+                env={"PEX_TOOLS": "1"},
+                check=True,
+                capture_output=True,
```
kaos, line 102: This silences the "Uninstalling setuptools..." logging sometimes observed which can cause issues when you expect structured output on stdout.

kaos:
Uh... still need the find-links thing for Pants <2.. 😬

tools/src/scie_pants/pants_version.py
```diff
@@ -206,3 +171,58 @@ def configure_version():
...
     )
+
+
+PYTHON_IDS = {
+    # N.B.: These values must match the lift TOML interpreter ids.
+    "cp38": "cpython38",
+    "cp39": "cpython39",
+    "cp310": "cpython310",
+    "cp311": "cpython311",
+}
```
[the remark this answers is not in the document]
kaos: Seems correct, based on the rejected alternative of `cp3_10` etc..
https://peps.python.org/pep-0641/

tools/src/scie_pants/install_pants.py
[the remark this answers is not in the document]
kaos: ... and restored again, to support using pants v1.x

tools/src/scie_pants/pants_version.py
```diff
@@ -23,16 +24,14 @@
 
 log = logging.getLogger(__name__)
 
-# After this version, Pants is released as a per-platform PEX using GitHub Release assets.
-# See https://github.com/pantsbuild/pants/pull/19450
-PANTS_PEX_GITHUB_RELEASE_VERSION = Version("2.18.0.dev5")
```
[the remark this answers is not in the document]
kaos: ehh.. all 2.x versions. So we adjust it down to `2.0.0` (should prob revert the inadvertent name change tho..)

cburroughs:
https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28

"The primary rate limit for unauthenticated requests is 60 requests per hour."

Just wanted to highlight the docs, I don't have a strong sense of how "risky" this is.

thejcannon:
> Uh... still need the find-links thing for Pants <2.. 😬 

Im happy to have a discussion about dropping v1 support. And telling people to use an older scie-pants for that.

thejcannon:
> https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28
> 
> "The primary rate limit for unauthenticated requests is 60 requests per hour."
> 
> Just wanted to highlight the docs, I don't have a strong sense of how "risky" this is.

What's this in relation to? If it's the PEX downloading that isn't the API, and isnt limited.

kaos:
> > Uh... still need the find-links thing for Pants <2.. 😬
> 
> Im happy to have a discussion about dropping v1 support. And telling people to use an older scie-pants for that.

I considered dropping v1 support as well. But then again, I don't feel it is motivated to drop it at this time, as it was working, and it didn't cost me much to preserve that functionality--I was just not aware of v1 being a thing still 😅 

If it becomes difficult to maintain, I'm +1 to dropping it in a future version. Users still on Pants v1 already show they're OK staying on an old version, so feels like freezing their version of supported scie-pants should be no worse off.

tools/src/scie_pants/pants_version.py
```diff
@@ -206,3 +175,72 @@ def configure_version():
...
+
+
+def fetch_github_api(ptex: Ptex, path: str, github_api_bearer_token: str | None = None) -> Any:
+    headers = (
+        {"Authorization": f"Bearer {github_api_bearer_token}"} if github_api_bearer_token else {}
+    )
+    return ptex.fetch_json(f"{GITHUB_API_BASE_URL}/{path}", **headers)
+
+
+def get_release_asset_and_python_id(
```
thejcannon, line 227: Instead of using the API, what if we just took shots in the dark?

Since this is during Pants bootstrapping, the timing of the "failed" requests wouldn't matter much.

So, e.g. just try to see if we can download the asset with Py312, otherwise try again with py 3.11, and so on (or whatever the Python versions would be).

kaos: That would work, of course. Do we need to? (i.e. is the api rate limit an actual obstacle here?)

There seems to be response headers we could examine to learn more, if we want to investigate further: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#checking-the-status-of-your-rate-limit

thejcannon: It just feels disingenuous to eat these on the users behalf without them knowing, since there are other ways.

thejcannon: Actually, wait.

We'd never release a newer version of pants that uses a lower Python than the previous...

And when we bump the Python version, we're going to bump scie-pants (to add support). That means:
- either we know the Python version given a pants version. It's in the known set, or it isn't and we haven't bumped python yet
- or we bumped python and scie-pants doesn't know it

Meaning that the code should:
- use the specific Python version if it recognizes (e.g. 2.16 uses py39, 2.20 uses py39)
- use the latest Python version (if scie-pants of a particular version is running, and the last hardcoded version mapping says that Pants 2.25 uses py311, try py311 for pants version 2.29. either it _does_ use it, then yay, or it doesn't and scie-pants should be upgraded)

Does that make sense?

kaos: > It just feels disingenuous to eat these on the users behalf without them knowing, since there are other ways.

That is a fair point.

> And when we bump the Python version, we're going to bump scie-pants (to add support).

Not sure I agree.. we can update scie-pants in advance to add the Python version not knowing at which time Pants will start using it.

I would prefer to have this de-coupled. We ship scie-pants with support for a number of Pythons.. and then it picks the correct one to use based on which version is needed for a particular version of Pants, which is not hard coded in scie-pants. Of course, historic versions could be hard coded to save on calls/time etc, and is likely the normal case most of the time, but it feels awkward to require a scie-pants release+upgrade whenever Pants bumps it's Python version.

thejcannon: I see.

Yeah OK, maybe we combine the best of both worlds. Try-if-unknown, but any time we bump Python for Pants we also update scie-pants (so it is therefore not unknown). That way we stop firing blind, but also don't _require_ people upgrade.

cburroughs: > Yeah OK, maybe we combine the best of both worlds

I like this idea.

src/main.rs
```diff
@@ -171,37 +171,11 @@ fn get_pants_process() -> Result<Process> {
             (None, None, None, false)
         };
 
-    let env_pants_sha = env_version("PANTS_SHA")?;
     let env_pants_version = env_version("PANTS_VERSION")?;
-    if let Some(pants_sha) = &env_pants_sha {
```
huonw, line 176: What do you think about landing this as a separate pull request? It'd at least help review by making this one smaller.

tools/src/scie_pants/pants_version.py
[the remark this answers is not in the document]
huonw: If we wanted to be extra precise, it could potentially be 2.0.0.dev0: https://github.com/pantsbuild/pants/releases/tag/release_2.0.0.dev0

huonw (commented):
Nice one. 

I'd find it a bit easier to review if we split up the changes:

- remove support for PANTS_SHA
- fix the stdout bug

Is that too annoying to do? Thanks

kaos:
> Nice one.
> 
> I'd find it a bit easier to review if we split up the changes:
> 
> * remove support for PANTS_SHA
> * fix the stdout bug
> 
> Is that too annoying to do? Thanks

No, that's a good move. It kind of fell in my lap while going for the Py support changes.. but breaking those out makes sense.

tools/src/scie_pants/pants_version.py
kaos: Yea, I'll rework this PR to break out tangential/unrelated changes per @huonw suggestion and fix the logic for determining Python version per @thejcannon suggestion.

Thanks for all reviews/comments so far :)

tools/src/scie_pants/install_pants.py
```diff
@@ -105,20 +105,10 @@ def install_pants_from_pex(
...
-                f"Wasn't able to fetch the Pants PEX at {pex_url}.\n\n{suggestion}"
-                "Check to see if the URL is reachable (i.e. GitHub isn't down) and if"
-                f" {pex_name} asset exists within the release."
-                " If the asset doesn't exist it may be that this platform isn't yet supported."
-                " If that's the case, please reach out on Slack: https://www.pantsbuild.org/docs/getting-help#slack"
-                " or file an issue on GitHub: https://github.com/pantsbuild/pants/issues/new/choose.\n\n"
-                f"Exception:\n\n{e}"
+                f"Wasn't able to fetch the Pants PEX at {pex_url}.\n"
+                "Check to see if the URL is reachable.\n\n"
+                f"Exception:\n{e}"
```
kaos, line 111: Simplified error message, as we shouldn't get here unless we have a working URL already..

tools/src/scie_pants/pants_release_tags.json
kaos, line 1: We'll never consult these tags for any of the 2.x versions..

tools/src/scie_pants/pants_version.py
kaos: I opted for building the expected URLs and doing a `HEAD` request to verify the Pants PEX exists as a way to determine what is available. We make an educated guess as to which Python version to use, failing that we try with all versions we support using.

i.e. There are no more Github API calls.

tools/src/scie_pants/pants_version.py
```diff
@@ -162,5 +182,120 @@ def configure_version():
...
+    if py:
+        pex_url = get_download_url(version, platform, py, ptex_urls)
+        if pex_url:
+            return pex_url, py
+
+    # Else, try all known Pythons...
+    for maybe_py in PYTHON_IDS.keys():
+        pex_url = get_download_url(version, platform, maybe_py, ptex_urls)
+        if pex_url:
+            return pex_url, maybe_py
```
huonw, line 256: For this fallback, we could take a different tack:

1. hit the GH API `/repos/pantsbuild/pants/releases/releasee_[version}/assets` (https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28#list-release-assets) to get a list of all assets
2. If not 200: if 404 => surface as unknown version error to user; if some other problem => surface that error more directly to users (e.g. connection error, 5xx from GitHub)
3. Look through the assets for the one matching the appropriate platform and extract the python identifier from that?

This likely simplifies the error reporting and does only a single request, rather than first trying cp38, then cp39, before eventually getting to the newer versions.

If the API urls have different rate limits to the one we're `HEAD`ing, this is potentially a problem... but I think we already pass around a `github_api_bearer_token` that gives users options, and we can also advise "please upgrade your scie-pants" since the new version is likely to have the URL included.

What do you think?

tools/src/scie_pants/pants_version.py
```diff
@@ -162,5 +182,120 @@ def configure_version():
...
+        pex_url = ptex_urls.get(pex_name)
+        if not pex_url:
+            return None
+    else:
+        pex_url = (
+            f"https://github.com/pantsbuild/pants/releases/download/release_{version}/{pex_name}"
+        )
+    req = urllib.request.Request(pex_url, method="HEAD")
+    try:
+        with urllib.request.urlopen(req) as rsp:
```
huonw, line 286: I think we should probably set a timeout here, e.g. 
```suggestion
        with urllib.request.urlopen(req, timeout=10) as rsp:
```

Potentially it should be configurable too?

tools/src/scie_pants/install_pants.py
```diff
@@ -76,19 +75,20 @@ def install_pants_from_pex(
...
     extra_requirements: Iterable[str],
-    find_links: str | None,
     bootstrap_urls_path: str | None,
 ) -> None:
     """Installs Pants into the venv using the platform-specific pre-built PEX."""
-    uname = os.uname()
-    pex_name = f"pants.{version}-cp39-{uname.sysname.lower()}_{uname.machine.lower()}.pex"
-
-    pex_url = f"https://github.com/pantsbuild/pants/releases/download/release_{version}/{pex_name}"
+    pex_name = pex_url.rsplit("/", 1)[-1]
```
huonw, line 84: Maybe we could use:

```suggestion
    pex_name = os.path.basename(pex_url)
```

tools/src/scie_pants/install_pants.py
```diff
@@ -76,19 +75,20 @@ def install_pants_from_pex(
...
     if bootstrap_urls_path:
```
huonw, line 85: I think potentially this whole codepath should be removed? We should just trust `pex_url` to have been computed by the appropriate bootstrap-URLs look-ups earlier and don't need to re-do it. When removing this, we should keep the error messages, i.e. they'll need to move somewhere else.

tools/src/scie_pants/pants_version.py
```diff
@@ -162,5 +182,120 @@ def configure_version():
...
+) -> tuple[str, str] | None:
+    if bootstrap_urls_path:
+        bootstrap_urls = json.loads(Path(bootstrap_urls_path).read_text())
+        ptex_urls = bootstrap_urls.get("ptex")
+        if ptex_urls is None:
+            raise ValueError(
+                f"Missing 'ptex' key in PANTS_BOOTSTRAP_URLS file: {bootstrap_urls_path}"
+            )
+    else:
+        ptex_urls = None
```
huonw, line 244: What do you think about putting all of the bootstrap URLs handling here, and then the rest of this function doesn't have to think about it? 

In particular, if someone specifies this file, the intent is "do all downloads from these URLs". If a file is missing, we should give an error about that, but currently this new code will fallback to the URL-probing (i.e. doing external requests) which is a bit surprising.

I think we can also "trust" those URLs and don't need to validate them, particularly, which eliminates the `file://` URL handling that was previously required.

So, maybe something like:

```suggestion
    if bootstrap_urls_path:
         bootstrap_urls = json.loads(Path(bootstrap_urls_path).read_text())
         ptex_urls = bootstrap_urls.get("ptex")
         if ptex_urls is None:
             raise ValueError(
                 f"Missing 'ptex' key in PANTS_BOOTSTRAP_URLS file: {bootstrap_urls_path}"
             )

         # do all the validation that was previously in `install_pants.py`
         return ...
```

And then remove the `ptex_urls` arg to `get_download_url`.

tools/src/scie_pants/pants_version.py
```diff
@@ -162,5 +182,120 @@ def configure_version():
...
+
+    py = get_python_id_for_pants_version(version)
+    if py:
+        pex_url = get_download_url(version, platform, py, ptex_urls)
```
huonw, line 248: Do we know if these URLs have rate limits for `HEAD` requests?

huonw (commented):
Exciting!

tools/src/scie_pants/pants_version.py
kaos: Unsure. It seems there could be a limit approx. 5k req/ip-address/hour.
https://stackoverflow.com/a/74960542/444060

tools/src/scie_pants/pants_version.py
```diff
@@ -162,5 +182,120 @@ def configure_version():
...
+    for maybe_py in PYTHON_IDS.keys():
+        pex_url = get_download_url(version, platform, maybe_py, ptex_urls)
+        if pex_url:
+            return pex_url, maybe_py
+
+    return None
+
+
+def get_python_id_for_pants_version(version: Version) -> str | None:
+    versions = json.loads(importlib.resources.read_text("scie_pants", "pants_python_versions.json"))
```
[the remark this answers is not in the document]
kaos: 🤷🏽 I wouldn't prefer that if the list would tend to get long, but I imagine it'll be years before it's even filling a single page on screen, and by then we may have refactored this to something else too.. so I'm fine either way.

tools/src/scie_pants/install_pants.py
```diff
@@ -166,12 +154,13 @@ def main() -> NoReturn:
...
     parser.add_argument(
-        "--pants-version", type=Version, required=True, help="The Pants version to install"
+        "--pants-version", type=Version, required=True, help="The Pants version to install."
     )
+    parser.add_argument("--pants-pex-url", type=str, help="The pants pex release asset url.")
     parser.add_argument(
         "--find-links",
         type=str,
-        help="The find links repo pointing to Pants pre-built wheels for the given Pants version",
+        help="The find links repo pointing to Pants pre-built wheels for the given Pants 1.x version.",
```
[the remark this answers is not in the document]
kaos: Even `0.x`, but I've no idea if those will actually run...???
https://github.com/pantsbuild/scie-pants/blob/main/tools/src/scie_pants/pants_release_tags.json

tools/src/scie_pants/pants_python_versions.json
```diff
@@ -0,0 +1,4 @@
+[
+  {"pants": "2.5.0.dev0", "python": "cp39"},
+  {"pants": "2.0.0.dev0", "python": "cp38"}
+]
```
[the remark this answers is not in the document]
kaos: Yea, add to *the top* of the list, and it would be `cp312` and we also need to add that to the list of cpython interpreters in the lift manifest and known python versions.

I'll add a section in the readme about how to add a new python to the list of supported pythons along with how to declare a new cpython version break for Pants.

tools/src/scie_pants/pants_version.py
```diff
@@ -162,5 +182,120 @@ def configure_version():
...
+            if rsp.status == 200:
+                return pex_url
+            if (
+                ptex_urls
+                and rsp.status is None
+                and pex_url.startswith("file://")
+                and rsp.headers.get("Content-Length")
+            ):
+                return pex_url
+    except Exception:
```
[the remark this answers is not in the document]
kaos: May even refactor this away, per Huon's feedback. But yea, more logging is def worth while to add in general. I'll keep that in mind, thanks :)

tools/src/scie_pants/configure_pants.py
```diff
@@ -63,6 +63,11 @@ def main() -> NoReturn:
     parser.add_argument(
         "--github-api-bearer-token", help="The GITHUB_TOKEN to use if running in CI context."
     )
+    parser.add_argument(
+        "--pants-bootstrap-urls",
```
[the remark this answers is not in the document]
kaos: It's a single file, yes, but that file may contain many URLs... 😄 so, I guess it's missing a "_file" suffix: `--pants-bootstrap-urls-file`. But as you said, the name's been set already, so I'm just rolling with it..

tools/src/scie_pants/pants_version.py
kaos: I'm about to dive back in here.. regarding hitting the GH api, that was in the previous iteration but was dropped in favor of just going with we "know" should work to avoid eating on the API rate limit.. for me, I feel either way is OK; I guess we just need to pick our poison here. (or, maybe go with a hybrid solution, perhaps we could do that iteratively though, to make it easier to reason about once we have a working flow in place..?)

tools/src/scie_pants/pants_version.py
kaos: > [...] but currently this new code will fallback to the URL-probing (i.e. doing external requests) which is a bit surprising.

I don't think this is correct. What it does is, it checks that the provided URL actually works before accepting it. If there is no URL provided for a particular platform/python we get `None` and proceed with trying all known Pythons etc, but `get_download_url` will never hit the network if you use the bootstrap urls file (for any other url than provided by that file).

tools/src/scie_pants/pants_version.py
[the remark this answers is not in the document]
kaos: inlining.. the list will likely be manageably short..

tools/src/scie_pants/pants_version.py
```diff
@@ -162,5 +182,120 @@ def configure_version():
...
+        pass
+
+    if ptex_urls:
+        fatal(f"Bad URL in PANTS_BOOTSTRAP_URLS for {pex_name}: {pex_url}\n\n")
```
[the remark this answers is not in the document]
kaos: rephrased this message as being an URL that we failed to check (i.e. unreachable or other.. the error is included in the message)

tools/src/scie_pants/pants_python_versions.json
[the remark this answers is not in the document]
kaos: inlined in the python code, with a comment.

tools/src/scie_pants/install_pants.py
kaos: moved all bootstrap urls path checking to pants_version, invoked during configure pants..

tools/src/scie_pants/pants_version.py
kaos: Using env var `PANTS_BOOTSTRAP_URL_REQUEST_TIMEOUT_SECONDS`, defaulting to `10`.

tools/src/scie_pants/pants_version.py
kaos: > rather than first trying cp38, then cp39, before eventually getting to the newer versions.

besides using the API (which I too feel could be the preferable alt), we could also list the pythons in reverse order, so we try with the newest python first, more likely to get a hit sooner.

huonw (approved):
Really nice!

I like it a lot... however, it's a moderately complicated switch. We could consider doing a pre-release release containing this change (e.g. 0.12.0rc0) that we can manually test for a day or two before doing a proper full release.

In particular, I think `get-pants.sh` without `--version` will automatically pick up the "latest" release. Thus, if people do an unpinned install in CI (highly likely), every time we do a release we instantly start running the new code in people's CI. This means if we've made a mistake here, we'll have a bit of a panic to get it fixed up. By doing a non-"latest" pre-release first we'll have a bit of breathing room to validate.

Alternatively, we can just hope for the best! What do you think?

kaos:
> Alternatively, we can just hope for the best! What do you think?

I'm in favor of doing a pre-release. Prefer to avoid stressful bugs :p

kaos:
@huonw I've added support for manually running the release workflow for a pre-release tag. The pre-release is not announced (besides to the GH discussions topic) nor published to our brew tap, so flies entirely under the radar. (had a bug so it wasn't marked as a pre-release properly, but I hope it was just my typo which is now fixed.) 

Only way to use it is to explicitly download the binary from GH release assets and run/install it that way.

I've made a quick smoke test on a repo of mine using the 0.12.0-beta.0 version on Mac M2, seems to be working fine there at least.

huonw:
Oh, cool!

My testing:

- Works locally for me too:
  | OS \ arch | x86-64 | arm64 |
  |---|---|---|
  | macOS | ✅ (native) | ✅  (native) |
  | Linux | ✅ (docker, emulated) | ✅  (docker) |
- I'm also running it through my work repo CI now (x86-64 Linux) seems to work fine too ✅ 

I feel like I have a pretty boring set-up though: everything connected to the internet always, no proxies/blocking. "Standard" operating systems etc. Do we want to solicit people to test more, especially if they have unusual set-ups? 

> Only way to use it is to explicitly download the binary from GH release assets and run/install it that way.

FYI looks like the `get-pants.sh` script we publish supports this with `./get-pants.sh --version 0.12.0-beta.0` too. 



> The pre-release is not announced

Minor process things, but I wonder if, for future, we could still announce these to Slack, similar to how we announce Pants dev releases. We can get this PR and its feature in, and iterate on the unrelated stuff once it's in, though.

kaos:
> Do we want to solicit people to test more, especially if they have unusual set-ups?

I'm willing to handle any edgy issues as they come up, as it seems to work fine for the general case it ought to work for most users. If there's too much issues, we can always roll back the latest version to 0.11.0 again to address them.



> FYI looks like the `get-pants.sh` script we publish supports this with `./get-pants.sh --version 0.12.0-beta.0` too.

Awesome.

> Minor process things, but I wonder if, for future, we could still announce these to Slack, similar to how we announce Pants dev releases. We can get this PR and its feature in, and iterate on the unrelated stuff once it's in, though.

Sounds good.

PR #375, opened 2024-03-13
Silence stdout/stderr during Pants setup.

kaos:
Closes pantsbuild/pants#20315
Broken out from #351

### DIFF
--- a/package/src/test.rs
+++ b/package/src/test.rs
@@ -29,8 +29,17 @@ macro_rules! integration_test {
     };
 }
 
-fn issue_link(issue: usize) -> String {
-    format!("https://github.com/pantsbuild/scie-pants/issues/{issue}")
+macro_rules! issue_link {
+    ($issue: expr) => {
+        issue_link($issue, "pantsbuild/scie-pants")
+    };
+    ($issue: expr, $repo: expr) => {
+        issue_link($issue, $repo)
+    };
+}
+
+fn issue_link(issue: usize, repo: &str) -> String {
+    format!("https://github.com/{repo}/issues/{issue}")
 }
 
 fn decode_output(output: Vec<u8>) -> Result<String> {
@@ -99,6 +108,7 @@ pub(crate) fn run_integration_tests(
         test_tools(scie_pants_scie, check);
         test_pants_bin_name_handling(scie_pants_scie);
         test_pants_bootstrap_handling(scie_pants_scie);
+        test_pants_bootstrap_stdout_silent(scie_pants_scie);
         test_tools_pex_reproducibility(workspace_root, tools_pex_path, tools_pex_mismatch_warn);
         test_pants_bootstrap_tools(scie_pants_scie);
 
@@ -238,7 +248,7 @@ fn test_pants_bin_name_handling(scie_pants_scie: &Path) {
         false,
         r#"
             [GLOBAL]
-            pants_version = "2.15.0rc5"
+            pants_version = "2.18.0"
             [anonymous-telemetry]
             enabled = false
             "#,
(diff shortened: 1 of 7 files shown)


tools/src/scie_pants/install_pants.py
```diff
@@ -120,23 +120,36 @@ def install_pants_from_pex(
...
+                    "--collisions-ok",
+                    "--no-emit-warnings",  # Silence `PEXWarning: You asked for --pip ...`
+                    "--disable-cache",
+                    str(venv_dir),
+                ],
+                env={"PEX_TOOLS": "1"},
+                check=True,
+                capture_output=True,
+            )
+            with open(str(venv_dir / "pants-install.log"), "a") as fp:
```
huonw, line 142: This currently incorporates both stdout and stderr into one file without separators and not with the content in chronological order. What do you think about either:
- writing to separate files 
- combining both streams into one so that the output is properly interleaved

>  If you wish to capture and combine both streams into one, use stdout=PIPE and stderr=STDOUT instead of capture_output. &mdash; https://docs.python.org/3/library/subprocess.html#subprocess.run

huonw (commented):
Thanks for splitting this up. In addition to the details of the file:

- how does a user know how to find this file, if they need to? Are there docs we could augment with a sentence? (There's the additional issue of the venv path being non-obvious with hashes in it, but that may be impossible to solve?)
- is this something we can add a test for do it doesn't regress?

tools/src/scie_pants/install_pants.py
kaos: Ah, good point. I think I'd prefer a single combined stream rather than two files. I'll push for that.

kaos:
> is this something we can add a test for so it doesn't regress?

I'm sure we could.. do you think it's worth looking into?

I assume it's the part about not leaking output to stdout during bootstrapping.. checking the presence of the `pants-install.log` file could also be done, but perhaps trickier as we'd need to find out the path.

In that case, it feels like we might prefer looking into adding support for something to query the lift runtime for the various resulting paths etc. perhaps something like `SCIE=lift.expand scie-pants scie.bindings`.

tools/src/scie_pants/install_pants.py
```diff
@@ -120,23 +120,37 @@ def install_pants_from_pex(
...
+                stderr=subprocess.STDOUT,
+            )
+        except subprocess.CalledProcessError as e:
+            fatal(
+                f"Failed to create Pants virtual environment.\nError: {e}, output:"
+                f"\n-----\n{e.stdout}\n-----\n"
+            )
+        else:
+            with open(str(venv_dir / "pants-install.log"), "a") as fp:
+                print(f"Installed {pex_name}", file=fp)
```
kaos, line 150: added extra log line to not just leave an empty file in case there were no additional output.

huonw:
> I'm sure we could.. do you think it's worth looking into?
>
> I assume it's the part about not leaking output to stdout during bootstrapping..

Yeah, just keeping `stdout` clean. My preference would be yes, but if it's a tonne of effort, maybe not worth it? 🤷

kaos:
> > I'm sure we could.. do you think it's worth looking into?
> > I assume it's the part about not leaking output to stdout during bootstrapping..
> 
> Yeah, just keeping `stdout` clean. My preference would be yes, but if it's a tonne of effort, maybe not worth it? 🤷

I had a long reply, which concluded that a test is worthwhile, as we want to ensure stdout is clean, regardless of what is being done during bootstrap, so it would protect not only against this fix in particular, but against all future changes that introduces new stuff during bootstrapping as well. I'll look into it :)

package/src/test.rs
```diff
@@ -238,7 +248,7 @@ fn test_pants_bin_name_handling(scie_pants_scie: &Path) {
         false,
         r#"
             [GLOBAL]
-            pants_version = "2.15.0rc5"
+            pants_version = "2.18.0"
```
kaos, line 251: By using the same pants version consistently, we avoid having to download and bootstrap more versions than necessary.

kaos:
@huonw OK, I've added a test verifying bootstrap doesn't pollute stdout. PTAL :)

huonw (approved):
Nice!

PR #384, opened 2024-04-03
Bump to pex 2.3.0

huonw:
General maintenance bump from 2.1.137 in Pants and 2.1.163 in pbt.toml to 2.3.0 in both places.

This will be required to run with Pants 2.20, which has a minimum supported PEX version of 2.1.148.

### DIFF
--- a/package/pbt.toml
+++ b/package/pbt.toml
@@ -20,8 +20,8 @@ version = "3.8.18"
 [[lift.files]]
 name = "pex"
 type = "blob"
-digest = { size = 3677552, fingerprint = "21cb16072357af4b1f4c4e91d2f4d3b00a0f6cc3b0470da65e7176bbac17ec35" }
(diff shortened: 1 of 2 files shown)


huonw:
The second commit is #385.

benjyw (approved):

huonw:
Hm, I don't know why macOS arm64 CI passed on #385, but now seems to be persistently failing on `main` and this PR. 😦

kaos (approved):

huonw:
> Hm, I don't know why macOS arm64 CI passed on https://github.com/pantsbuild/scie-pants/pull/385, but now seems to be persistently failing on main and this PR. 😦

Symptoms (of native code for `_setproctitle` being x86-64 not arm64, on Apple Silicon) make me assume that this was the same as https://github.com/pantsbuild/pants/issues/20759, which we've resolved/worked around via #388.

PR #385, opened 2024-04-03
Fix CI: Use a tmpdir SCIE_BASE to force bootstrapping, incorporate Python path into cache key

### LINKED ISSUE #270: CI failures when GitHub Actions upgrades its python patch version
We cache SCIE_PANTS_DEV_CACHE, which includes venvs that include symlinks into `/opt/hostedtoolcache/` (e.g., 

```
/home/runner/work/scie-pants/scie-pants/.scie_pants_dev_cache/venvs/pants-2.14.1/py39.venv/bin/python3.9 ->
 /opt/hostedtoolcache/Python/3.9.17/x64/bin/python3.9
```

However that symlink target may not exist on a new runner, e.g., when GHA upgrade their "Python 3.9" from 3.9.17 to 3.9.18.

huonw:
There were two issues with CI:

- It seems like the test that verifies silence on stdout wasn't always doing the required bootstrap. This makes the bootstrap always happen by forcing it to be starting with an empty cache (`SCIE_BASE`) by setting that to a fresh temp dir.
- The ubuntu runner was failing to run the `test_pants_source_mode` test, because the venv from the restored cache had an symlink to `/opt/hostedtoolcache/Python/3.9.18/x64/bin/python` (NB version 3.9.18), but the runners have updated to 3.9.19. Fixes #270.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,13 +53,13 @@ jobs:
         id: build_it_cache_key
         run: |
           # The caches include venvs which have absolute links to Python binaries, so our system
-          # should be resilient to this
+          # should be resilient to this (see `test_pants_source_mode` in `test.rs`).
           echo "cache_key=${{ matrix.os }}-scie-pants-v7-$(which python)" | tee -a "$GITHUB_OUTPUT"
       - name: Cache Build and IT Artifacts
         uses: actions/cache@v3
         with:
           path: ${{ env.SCIE_PANTS_DEV_CACHE }}
-          key: ${{ steps.build_it_cache_key.cache_key }}
+          key: ${{ steps.build_it_cache_key.outputs.cache_key }}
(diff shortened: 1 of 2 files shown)


huonw:
The failure seems to be unrelated. I haven't had time to look into it, yet.

```
fatal: No names found, cannot describe anything.
/home/runner/work/scie-pants/scie-pants/.scie_pants_dev_cache/clones/pants-2.14.1/pants: line 69: /home/runner/work/scie-pants/scie-pants/.scie_pants_dev_cache/venvs/pants-2.14.1/py39.venv/bin/python: No such file or directory
```

benjyw (approved):

huonw:
I've added another fix beyond the original change to set `SCIE_BASE`, because CI was still failing.

kaos (commented):
Thanks 😅

PR #386, opened 2024-04-04
Attempt to debug ubuntu failure

huonw:
Maybe it's a broken cache?


huonw:
Ah, this is the classic "cached a venv with symlink to python binary that doesn't exist on new runner". Specifically the cached venv has a symlink to `/opt/hostedtoolcache/Python/3.9.18/x64/bin/python3.9` but the new runners are getting `3.9.19` (patch version 19, not 18). 💥 

Will add the fix to #385.

PR #388, opened 2024-04-15
Use 2.21.0.dev6 for PANTS_SOURCE testing, not 2.14.1

huonw:
This updates the hard-coded version that's used for the `PANTS_SOURCE` test from 2.14.1 to 2.21.0.dev6, which is the current most recent `main` release.

This is acutely motivated by the old version using dependencies without appropriate wheels for macOS arm64 in particular, which causes CI to hit https://github.com/pantsbuild/pants/issues/20790, thus making development here slower and dependent on fixing those infrastructure issues.

It potentially also brings the Rust version used by that dependency to be closer to that used by the scie-pants repo, thus making Rust upgrades potentially easier (less juggling of rust versions required https://github.com/pantsbuild/scie-pants/pull/358).

Finally, the risk of this change is that we might regress `PANTS_SOURCE` behaviour for old Pants versions, e.g. someone is using a local check out of a 2.14.x-era release, and some future change to `scie-pants` breaks this. I think we can treat this sort of out-of-date `PANTS_SOURCE`s as semi-supported: if someone is affected, they can file an issue and we (potentially) fix, and they have the easy work-around of pinning to a lower/previously-working version of `scie-pants`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' || !startsWith(github.ref, 'refs/tags/') }}
 env:
   CARGO_TERM_COLOR: always
+  SCIE_PANTS_DEV_CACHE: ./.fixme-temporary-cache
 jobs:
   org-check:
     name: Check GitHub Organization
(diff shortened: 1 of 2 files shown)


kaos (approved):
Great! 🚀

PR #391, opened 2024-04-24
Use `ureq` for "science" download, not `cargo install`'d `ptex`

huonw:
This replaces an inline `cargo install` of https://github.com/a-scie/ptex with the `ureq` library. The `ptex` binary seems to only be used as a "portable" `curl -O <url>`, which can, I think, be done with code built into the `package/` lib itself.

This change:
- conceptually simplifies the "build scie-pants" process
- makes it more reliable/secure (I think the `cargo install` wasn't using `ptex`'s lockfile, because it didn't pass `--locked`)
- potentially speeds up the build when not cached (building ptex has to build a whole lot of libraries, taking 30+s for me).

It also gets rid of two explicit dependencies (`once_cell`, `url`)... but `ureq` depends on both, so they don't actually disappear from the dependency tree. 🤷 

I chose `ureq` because it doesn't seem to add much compilation time. With some benchmarks on my 4-core laptop:

| build type | time before (s) | time after (s) |
|---|---|---|
| clean | 9-10 | 10-11 |
| cached, comment change in `package/src/utils/build.rs` | 0.5-0.6 | 0.5-0.6 |

There's three commits, which can be reviewed individually. The middle one is the interesting one, with the other two minor related refactors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,11 +55,11 @@ jobs:
           # The caches include venvs which have absolute links to Python binaries, so our system
           # should be resilient to this (see `test_pants_source_mode` in `test.rs`).
           echo "cache_key=${{ matrix.os }}-scie-pants-v7-$(which python)" | tee -a "$GITHUB_OUTPUT"
-      - name: Cache Build and IT Artifacts
-        uses: actions/cache@v3
-        with:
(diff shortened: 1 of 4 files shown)


kaos (approved):

huonw:
I did a CI build without the cache, because the cached build wasn't exercising this new code path. Seems to work ✅

PR #420, opened 2024-11-04
Upgrade to pants 2.22.1

clintonsteiner:

PR #423, opened 2024-11-22
CI: update python version, actions versions, and move to latest opera…

clintonsteiner:
…ting systems


huonw:
I think this was an attempt to see if it helped with #420. Thanks for trying @clintonsteiner 

I'll reopen and run CI so you can see the effect of the changes and see if that informs your debugging.

Issue #424, opened 2024-11-21
Unable to run pants 2.25.0.dev0 due to "No interpreter compatible with the requested constraints was found:" error


huonw:
**Describe the bug**

Running pants 2.25.0.dev0 upgrades the version of Python used, but this doesn't seem to have worked completely, actually attempting to run Pants gives an error like:

```
Failed to find compatible interpreter on path /Users/huon/Library/Caches/nce/7af7058f7c268b4d87ed7e08c2c7844ef8460863b3e679db3afdce8bb1eedfae/cpython-3.11.9+20240415-aarch64-apple-darwin-install_only.tar.gz/python/bin/python3.11.

Examined the following interpreters:
1.) /Users/huon/Library/Caches/nce/7af7058f7c268b4d87ed7e08c2c7844ef8460863b3e679db3afdce8bb1eedfae/cpython-3.11.9+20240415-aarch64-apple-darwin-install_only.tar.gz/python/bin/python3.11 CPython==3.11.9

No interpreter compatible with the requested constraints was found:

  Version matches CPython<3.10,>=3.8
Error: Failed to establish atomic directory /Users/huon/Library/Caches/nce/6faa4322d1df41d032e4938795c6f2c262ab92bb642a9bac1101cb7d1631f9c1/locks/install-b4ff91a6f4dedb37b388e53c0e15b581f82e68706a77a04f1d9940d3e97e0da1. Population of work directory failed: Boot binding command failed: exit status: 1

Isolates your Pants from the elements.

Please select from the following boot commands:

<default> (when SCIE_BOOT is not set in the environment)  Detects the current Pants installation and launches it.
bootstrap-tools                                           Introspection tools for the Pants bootstrap process.
update                                                    Update scie-pants.

You can select a boot command by setting the SCIE_BOOT environment variable.
```

(Filed here not scie-pants because I'm guessing this might be some metadata in the Pants wheel? It seems like the interpreter is configured correctly.)

Reproducer: run `PANTS_VERSION=2.25.0.dev0 pants version`
                                                                                                                                              
**Pants version**
2.25.0.dev0

**OS**

macOS, Linux

**Additional info**
N/A

PR #425, opened 2024-11-25
Update tools.pex dependencies

huonw:
As preparation for #424, this just updates all of the Python dependencies of tools.pex to their latest versions.

This is a separate PR to #426 so that we can bisect and more easily tell the difference between "problem caused by dependency upgrade" vs. "problem caused by ICs change", if there are problems.

### DIFF
--- a/tools/lock.json
+++ b/tools/lock.json
@@ -24,37 +24,56 @@
           "artifacts": [
             {
               "algorithm": "sha256",
-              "hash": "99b87a485a5820b23b879f04c2305b44b951b502fd64be915879d77a7e8fc6f1",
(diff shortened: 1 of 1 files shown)


benjyw (approved):

huonw:
Hm, actually, maybe it does work... CircleCI ran in #428, and I can see things in the circle CI UI now: https://app.circleci.com/pipelines/github/pantsbuild/scie-pants

Maybe logging in at all was enough to re-invigorate it... I'll attempt to retrigger here.

huonw:
Yep, got the ✅s now. Thanks for listening.

PR #426, opened 2024-11-25
Fix support for Pants 2.25.x / Python 3.11 

### LINKED ISSUE #424: Unable to run pants 2.25.0.dev0 due to "No interpreter compatible with the requested constraints was found:" error

huonw:
Fixes #424 

This makes a few adjustments to support running Pants 2.25.0.dev0 and newer, which run using Python 3.11 (https://github.com/pantsbuild/pants/pull/21528), iterating on #351. Particularly:

- expanding tools.pex's interpreter constraints to include 3.11 (and also 3.10, which seems fine)
- add a test (NB. Pants 2.25 cannot run on macOS older than 13 / 14, depending on platform, so this test is conditional: https://github.com/pantsbuild/pants/pull/21655)
- hardcoding that 2.25.0.dev0 uses Python 3.11, to use the optimised "hit the exact URL first time" codepath, instead of having to try all Python versions: https://github.com/pantsbuild/scie-pants/blob/d50bd33fd67e944384c9548811a66c1cb03113a5/tools/src/scie_pants/pants_version.py#L239-L252

This thus preps scie-pants release 0.12.1 too, which I'll tag once merged.

### DIFF
--- a/package/src/test.rs
+++ b/package/src/test.rs
@@ -46,6 +46,43 @@ fn decode_output(output: Vec<u8>) -> Result<String> {
     String::from_utf8(output).context("Failed to decode Pants output.")
 }
 
+/// Returns true if the current platform is a macOS major version that's older than the requested minimums.
+///
+/// (NB. Running on a non-macOS platform will always return false.)
+fn is_macos_thats_too_old(minimum_x86_64: i64, minimum_arm64: i64) -> bool {
+    let min_major = match *CURRENT_PLATFORM {
+        Platform::MacOSX86_64 => minimum_x86_64,
+        Platform::MacOSAarch64 => minimum_arm64,
+        _ => return false,
+    };
+
+    let version_output = execute(
+        Command::new("sw_vers")
+            .arg("-productVersion")
+            .stdout(Stdio::piped()),
+    )
+    .unwrap();
+    let version_str = decode_output(version_output.stdout).unwrap();
+
+    // for this constrained use case, we can just parse the first element, e.g. 10.14 & 10.15 => 10,
+    // 11.0.1 => 11, etc.
+    //
+    // If the distinction between the 10.x "major" versions ends up mattering, feel free to refactor
+    // this to work with the full version string.
+    let major: i64 = version_str
+        .trim()
+        .split('.')
+        .next()
+        .and_then(|s| s.parse().ok())
+        .unwrap_or_else(|| {
+            panic!(
+                "Failed to parse macOS version from `sw_vers -productVersion` output: {}",
+                version_str
+            )
+        });
+    major < min_major
+}
+
 enum ExpectedResult {
     Success,
     Failure,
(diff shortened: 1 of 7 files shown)


package/src/test.rs
```diff
@@ -334,6 +372,31 @@ fn test_pants_bootstrap_tools(scie_pants_scie: &Path) {
     .unwrap();
 }
 
+fn test_pants_2_25_using_python_3_11(scie_pants_scie: &Path) {
+    integration_test!("Verifying we can run Pants 2.25+, which uses Python 3.11");
+    if is_macos_thats_too_old(13, 14) {
+        log!(
+            Color::Yellow,
+            "Pants 2.25 cannot run on this version of macOS => skipping"
+        );
```
kaos, line 381: It would be nice, if not in the message, but at least as a code comment, the reason we can't run pants 2.25 on these older versions of MacOS.

kaos (approved):

cburroughs:
Thanks for fixing this up.  I was able to build this and run not just `version` but `fmt`, `test`, etc.

cburroughs (approved):

PR #427, opened 2024-11-25
Use Pants'-managed runs-on runners for Linux aarch64 CI

huonw:
This replaces our use of Circle CI's arm64 Linux runners for both testing and releasing with the runs-on runners that Pantsbuild configures (and pays for) in AWS, via https://runs-on.com

This has a few benefits:

- single tool to learn for configuring CI
- single source of credentials (currently the release build is failing due to a token seemingly being invalid, but it's not obvious to me how to fix this)
- more atomicity, e.g. instead of the release needing to trigger a separate Circle CI build that runs in the background, we can just use the normal GitHub "wait for another job" rules

It also has downsides: we're going from free resources to paid. However, the incremental cost for the scie-pants repo should be quite small, compared to the main pants repo: 

- this repo is low volume (over the last 12 months we've done about 220 CI runs total)
- things run quickly (in this PR, ~7 minutes without a cache, ~4 minutes with)

I estimated cost based on US$0.1344/hour for a on-demand t4g.xlarge 4 core instance in us-east-1 (I'm not sure this is our exact configuration). Prices in practice may be cheaper, because Runs-On attempts to use [spot instances](https://runs-on.com/configuration/spot-instances/)).

- per build, we're looking at US$0.0090 (with cache) to US$0.016 (without cache)
- per month, we're looking at ~20 builds, so US$0.18 to US$0.32 (but likely cheaper given this is on-demand costs)

If this starts adding up a lot, we can likely optimise. For instance: optimise spot-instance selection for lowest cost not "PCO", and use more caching (whether that's coarse-grained `actions/cache` or fine-grained `sccache`).

Fixes #422, fixes #430

Issue #422, opened 2024-11-18
Replace CircleCI ARM CI with Pantsbuild's runs-on runners


huonw:
The main pants repo now uses "self-hosted" aarch64 linux runners, via runs-on and Pantsbuild's AWS account.

- https://github.com/pantsbuild/pants/pull/21325
- https://github.com/pantsbuild/pants/pull/21329
- https://github.com/pantsbuild/pants/pull/21330
- https://github.com/pantsbuild/pants/pull/21332

We can hopefully do the same here and simplify our configuration by only using Github Actions for CI instead of circle CI too.

We may want to move the pants repo config to be shared: https://runs-on.com/configuration/repo-config/#sharing-configuration-across-repositories

jsirois:
You're probably already aware, but another option is Docker QEMU emulation. I was gunning for $0 when I chose Circle CI and its great for that - I use it for a-scie projects for that reason - but another way to spend $0 at the sacrifice of speed is the emulation.  Examples here: https://github.com/pex-tool/pex/blob/a346870722dfff91cb335e77b6e6c44503e7f733/.github/workflows/gen-scie-platforms.yml#L60-L74

huonw:
Thanks; yeah, I considered it, but decided the incremental cost here is likely to be tiny, compared to spend on these runners in the (already-configured/already-in-use) CI in the main pants repo

(Analysis in https://github.com/pantsbuild/scie-pants/pull/427)

PR #428, opened 2024-11-25
Add runs-on config, inheriting from Pants, for Linux arm64 CI

huonw:
Preparation for #422: this add runs-on configuration to this repo, so we can using it to run GitHub Actions, replacing our CircleCI config (#427).

### DIFF
--- a/.github/runs-on.yml
+++ b/.github/runs-on.yml
@@ -1,0 +1,2 @@
+# reuse the same runner configuration as the main Pants repo
+_extends: pants


huonw:
I don't know if there's additional configuration we need to do here? Potentially we need to have the GitHub App registered for this repo too? https://runs-on.com/guides/install/#3-github-app-registration

huonw:
(Focusing on #426 / 0.12.1 before getting back to this.)

.github/runs-on.yml
```diff
@@ -0,0 +1,2 @@
+# reuse the same runner configuration as the main Pants repo
+_extends: pants
```
benjyw, line 2: Didn't know this was a thing, but nice!

benjyw (approved):

PR #439, opened 2024-12-09
Fix new draft-release release flow: avoid duplicates, capture changelog

huonw:
As observed in https://github.com/pantsbuild/scie-pants/pull/432#issuecomment-2526598191, after #432, release CI has two problems:

- It creates two releases, one as a draft that the artifacts are uploaded too, and then one that's published. I'm guessing this is because the `name` key wasn't set on the draft uploads
- The release notes couldn't be created, because the repo and its code (particularly `CHANGES.md`) wasn't available in the publishing step


benjyw (approved):

huonw:
This fixes the changelog issue but not the double-release-creation one. 😭 

Clearly I'm not understanding something here and need to dig deeper.

image

Issue #421, opened 2024-11-18
CI uses pantsbuild's self-hosted macOS runners


huonw:
Currently CI uses the org's self-hosted (in macstadium) macOS runners (macOS-10.15-X64, macOS-11-ARM64), but we'd prefer to not have to pay for and maintain the machines if we can. 

We should:

- replace the usage with Github-hosted runners if we can
- attempt to do a nice deprecation phasing (maybe including on the Pants-side, too)

Refs:

https://github.com/pantsbuild/scie-pants/blob/d50bd33fd67e944384c9548811a66c1cb03113a5/.github/workflows/ci.yml#L31

https://github.com/pantsbuild/scie-pants/blob/d50bd33fd67e944384c9548811a66c1cb03113a5/.github/workflows/release.yml#L62

- https://github.com/pantsbuild/pants/pull/21655
- https://github.com/pantsbuild/pants/issues/21413
- https://github.com/pantsbuild/pants/issues/21074

sureshjoshi:
Done in #519

PR #440, opened 2024-12-09
Build on macos-13, run on 10.15

huonw:
Some experiments for #421: does scie-pants built on newer macOS run on older versions?


benjyw:
Superseded by https://github.com/pantsbuild/scie-pants/pull/458?

huonw:
yeah, seems so

PR #441, opened 2024-12-09
Build on newer-versions of Github-hosted macOS runners

### LINKED ISSUE #421: CI uses pantsbuild's self-hosted macOS runners

huonw:
This switches us to building macOS binaries on Github-hosted runners:

- macOS 13 for x86-64
- macOS 14 for arm64

This then follows https://rust-cli.github.io/book/tutorial/packaging.html#building-binary-releases-on-ci to set `MACOSX_DEPLOYMENT_TARGET` to preserve our platform support.

Fixes #421

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,18 +34,23 @@ jobs:
         include:
           - os: ubuntu-22.04
             name: ubuntu-22.04
+            requires-setup-python: true
 
           - os: [runs-on, runner=4cpu-linux-arm64, image=ubuntu22-full-arm64-python3.7-3.13, "run-id=${{ github.run_id }}"]
             name: linux-arm64
+            requires-setup-python: false
 
           - os: macos-13
             name: "macos-13 x86-64"
+            requires-setup-python: true
 
           - os: macos-14
             name: "macos-14 arm64"
+            requires-setup-python: true
 
           - os: windows-2022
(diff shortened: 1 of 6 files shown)


huonw:
This keeps getting stuck on seemingly unrelated CI issues, e.g. the science invocations were hitting GitHub rate limits for downloading PBS releases... but why does that occur in this PR and not in all the other builds we do?? Dno.

benjyw:
Superseded by https://github.com/pantsbuild/scie-pants/pull/458

PR #444, opened 2024-12-11
Add workaround for release publishing action mishandling `draft: true` -> `false` transition

huonw:
This works around `softprops/action-gh-release` not being able to properly "publish" a release by changing `draft: true` to `draft: false`.

The intention is that the step with `draft: false` would find the existing draft release and update it (publishing it). This is what correctly happened for a testing `v0.12.3-beta.3` release from this branch: https://github.com/pantsbuild/scie-pants/releases/tag/v0.12.3-beta.3 (I also did attempted a `v0.12.3-beta.2` from this branch, but that used a newer version of the action code and failed completely, see below.)

Previously, the action would fail to find the existing draft release and instead create a duplicate one. The old behaviour is visible (to people who can see draft releases) with `v0.12.3-beta.0` and `-beta.1`:

image

The particular new version of the action is:

- changes: https://github.com/softprops/action-gh-release/compare/master...huonw:action-gh-release:huonw/pantsbuild-fix-see-scie-pants-439 (specifically commit https://github.com/huonw/action-gh-release/commit/998f80d5380609557d7464b01d59a10d845600a0 from branch https://github.com/huonw/action-gh-release/commits/huonw/pantsbuild-fix-see-scie-pants-439 )
- this is v2.0.9 (the current `v2` tag) with https://github.com/softprops/action-gh-release/pull/316 merged in
- using v2.2.0 (the latest version of the action) fails while uploading artifacts with errors like `Request body length does not match content-length header`, so this sticks with the older version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -97,7 +97,10 @@ jobs:
 
       # Build up a draft release with the artifacts from each of these jobs:
       - name: Create ${{ needs.determine-tag.outputs.release-tag }} Release
-        uses: softprops/action-gh-release@v2
+        # Need to pull in https://github.com/softprops/action-gh-release/pull/316 to work-around
+        # double-release-creation that happens when attempting to update a draft release to
+        # not-draft.
+        uses: huonw/action-gh-release@55ec297e11bf84ff0f5893a84ea987d93489920c # v2.2.0 + https://github.com/softprops/action-gh-release/pull/316
         env:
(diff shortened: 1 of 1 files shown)


benjyw (approved):
Hopefully that PR gets upstreamed soon... Can we prompt them about it?

huonw:
I've left a comment because it seemed like seeing real world usage was part of the review process, good call-out.

PR #459, opened 2025-05-16
Upgrade to Pants 2.26.0 and Pex 2.30.0.

benjyw:


pants.toml
```diff
@@ -20,7 +17,7 @@ config = "tools/pyproject.toml"
 
 [python]
 enable_resolves = true
-interpreter_constraints = [">=3.8,<3.10"]
+interpreter_constraints = [">=3.8,<3.12"]
```
benjyw, line 20: Necessary so that the tools.pex's compare equal

package/src/tools_pex.rs
```diff
@@ -58,7 +58,7 @@ pub(crate) fn build_tools_pex(
                     "--style",
                     "universal",
                     "--pip-version",
-                    "22.3",
+                    "24.3.1",
```
sureshjoshi, line 61: isnt pip on 25.x?

pants.toml
sureshjoshi: Any reason why it's 3.8? Can this be bumped?

package/src/tools_pex.rs
benjyw: Yeah, not trying for the latest and greatest, just something less archaic.

PR #478, opened 2025-07-15
Fix MacOS matrix references in CI.

benjyw:
- In ci.yml incorrect labels were referenced.
- In release.yml incorrect labels were requested.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,11 +66,11 @@ jobs:
           - os: [runs-on, runner=4cpu-linux-arm64, image=ubuntu22-full-arm64-python3.7-3.13, "run-id=${{ github.run_id }}"]
             name: linux-arm64
 
-          - os: macOS-13-X64
-            name: macOS-13-X64
+          - os: macos-13
+            name: macos-13-x86
 
-          - os: macOS-14-ARM64
-            name: macOS-14-ARM64
+          - os: macos-14
+            name: macos-14-arm64
     environment: Release
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -66,11 +66,11 @@ jobs:
           - os: [runs-on, runner=4cpu-linux-arm64, image=ubuntu22-full-arm64-python3.7-3.13, "run-id=${{ github.run_id }}"]
             name: linux-arm64
 
-          - os: macOS-13-X64
-            name: macOS-13-X64
+          - os: macos-13
```
sureshjoshi, line 69: Some confusion for me still is. We "run-on" the `os` item, which is `macos-13` and `macos-14`... How do we know which is arm, which is x86?

I thought Github named the runner `macos-14-arm`?

sureshjoshi (approved):

.github/workflows/release.yml
sureshjoshi: I mean, the arm one is running on the correct OS - but I just have no idea where the x86 and arm64 conventions come from.

```
Current runner version: '2.326.0'
Operating System
  macOS
  14.7.6
  23H626
Runner Image
  Image: macos-14-arm64
  Version: 20250709.1652
  Included Software: https://github.com/actions/runner-images/blob/macos-14-arm64/20250709.1652/images/macos/macos-14-arm64-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/macos-14-arm64%2F20250709.1652
```

benjyw: I think they just come from the list you linked to. This is why we add `name`, to clarify in the file.

PR #492, opened 2025-11-25
preserve environ (esp HOME) when calling scie in subprocess

### LINKED ISSUE #263: `SCIE_BOOT=update pants` results in installation error
Attempting to upgrade the `pants` launcher binary on my computer results in an installation error. Full output and log file can be found below.

```
jtilahun@JTN86G3:~/devel/monorepo$ SCIE_BOOT=update pants
Error: Isolates your Pants from the elements.

scie-pants
bootstrap-tools
pants
pants-debug
update

You can select a boot command by passing it as the 1st argument or else by setting the SCIE_BOOT environment variable.

ERROR: Failed to expand home dir in path ~/.nce
Install failed: Command '['/home/jtilahun/tools/bin/pants']' returned non-zero exit status 1.
More information can be found in the log at: /home/jtilahun/.cache/nce/65aa4f2a6c1f9bac672c0df94ae34c7170e5c071cda35e9b725945831905c122/bindings/logs/record-scie-pants-info.log

Error: Isolates your Pants from the elements.

scie-pants
bootstrap-tools
pants
pants-debug
update

You can select a boot command by passing it as the 1st argument or else by setting the SCIE_BOOT environment variable.

ERROR: Failed to establish atomic directory /home/jtilahun/.cache/nce/65aa4f2a6c1f9bac672c0df94ae34c7170e5c071cda35e9b725945831905c122/locks/scie-pants-info-6a8ddab0f3a22eb4476a6f164c2267a4abc996c084396b7b224489e041c4e369. Population of work directory failed: Boot binding command failed: exit status: 1
```

[record-scie-pants-info.log](https://github.com/pantsbuild/scie-pants/files/12481030/record-scie-pants-info.log)

cognifloyd:
Fixes #263

#13 introduced both of these `PANTS_BOOTSTRAP_VERSION=report pants` subprocess calls with one crucial difference: one of them does not preserve `os.environ`. Without that, scie failes to find its nce cache, and everything falls over.

This one preserves `os.environ`:
https://github.com/pantsbuild/scie-pants/blob/e61e619599983ac4f6efd4fe627ec70f725d58ea/tools/src/scie_pants/update_scie_pants.py#L184-L186

This one does not preserve `os.environ` causing the failure documented in #263.
https://github.com/pantsbuild/scie-pants/blob/e61e619599983ac4f6efd4fe627ec70f725d58ea/tools/src/scie_pants/record_scie_pants_info.py#L37-L39

Why does this happen? When a host depends on NSS for the `passwd` database (eg for remote AD or LDAP accounts made available to the system through SSSD), scies cannot resolve `~` without the `HOME` env var. Ultimately, this happens because scies include a pre-compiled a-scie/jump binary built against musl libc instead of glibc (a-scie/jump releases for linux-x86_64 and linux-aarch64 are always compiled with musl libc). To find the nce cache, a-scie/jump uses the `dirs` lib to look up the home directory by first checking for the `HOME` env var, and falling back to querying libc. This falls over because musl linux does not support NSS, so the fallback to musl libc `getpwuid_r` calls always fail for accounts provided by NSS.

If the `SCIE_BASE` env var was set, then a-scie/jump would not need to look up the home directory, but that still would not help without actually preserving the env vars when running the scie. So, we need to make sure at least `HOME` and any `SCIE_*` vars get passed to any subprocess that runs a scie. This PR just passes all of env vars, which aligns the two `PANTS_BOOTSTRAP_VERSION=report pants` calls so they are doing the same thing.

### DIFF
--- a/tools/src/scie_pants/record_scie_pants_info.py
+++ b/tools/src/scie_pants/record_scie_pants_info.py
@@ -36,7 +36,7 @@ def main() -> NoReturn:
 
     version = subprocess.run(
         args=[options.scie],
-        env={"PANTS_BOOTSTRAP_VERSION": "report"},
+        env={**os.environ, "PANTS_BOOTSTRAP_VERSION": "report"},
         stdout=subprocess.PIPE,
         text=True,
         check=True,


cognifloyd:
Is there a way to mock the libc `getpwuid_r` function used by the scie subprocess? 😵‍💫

Maybe if we can inject a bogus UID in the test, relying on the behavior of dirs checking `HOME` before falling back to the libc `getpwuid_r`? So, w/ bogus UID or USER, it should fail without a HOME env var, but pass with it? I'll play with that and see if I can get it to fail.

cognifloyd:
> Eeesh, I'm absolutely not a mock fan.
> 
> > In some cases, the rust dirs package cannot resolve ~ without the HOME env var
> 
> I'd like to be able to better enumerate that, so we could create that situation. Or, force that situation. e.g. could we run in a container or `lxc` or something and force this problem?
> 
> It may not matter long term, though, as the hope is to move away from this mechanism in favour of a simpler per-release scie (where we then might have to have similar functionality, though).

Here's a write up on why this happens (jump uses musl which does not support nss, so it can't find the AD/LDAP accounts that are only available via NSS + SSSD): https://github.com/pantsbuild/scie-pants/issues/263#issuecomment-3584325901

I've spent far too much time on this now. Adding an automated test for this would be very convoluted (maybe configure nss in a container to use ldap in another container and then?), and I really don't care to spend the time doing that. Do we really need all of that? If so, would you be willing to add it (feel free to push to my branch).

jsirois:
FWIW, `scie-jump`, `science` and Pex all now support glibc linked scie tips on Linux aarch64 and x86_64: https://github.com/pantsbuild/scie-pants/issues/263#issuecomment-3604298563

PR #502, opened 2026-02-25
Remove support for macos-13 intel.

benjyw:
Pants itself no longer runs on this platform, and
the github runner for it became unavailable as of
December 2025.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Release Notes
 
-## 0.12.6
+## 0.13.0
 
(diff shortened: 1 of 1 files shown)


sureshjoshi:
Nice! I've been messing around with `zig-cc` for some of my C projects, and it works really well for my trivial applications.

It was also [kicked around in the Rust threads](https://internals.rust-lang.org/t/bundle-zig-cc-in-rustup-by-default/22096) to bundle zig-cc with Rust (or solve the underlying problem via rust-centric means).

Glad to know cargo-zigbuild is as good as I'd heard about. We should absolutely kick the tires on it here and for the Pants engine.

PR #519, opened 2026-04-06
Using GitHub runners and removing Docker

sureshjoshi:
This PR consolidates the CI and release process, such that "release" now calls "ci" for output and all compilation/testing occurs on a single workflow action, rather than partial duplication.

As well, we only use GitHub Actions runners (i.e. `linux-arm`) so we can remove some temporary workarounds that were in-place for persistent workers, and remove the `runsOn` file.

Several external re-usable actions have been replaced with either the `gh` cli command, or more limited bash scripts.

A notable change is the removal of any `push`-based building for releases, in favour of workflow dispatches, directly based off `main`. This removes the need for any local tag signing.

### --- Original in-progress description ---

This is a mix of several different changes, which I'm playing with on my own fork (https://github.com/sureshjoshi/scie-pants/actions/runs/24041142971).

I'd like to consolidate the GHA and try to not have special cases if we can avoid it. I'd also like to avoid mixing Docker steps with the rest.

So, this PR attempts to do all that. I have a matrix without per-OS variants. I haven't tried running these yet, but on my fork, CI passes, and release is .... weird: https://github.com/sureshjoshi/scie-pants/deployments

I have some follow-up improvements to make, to not use unnecessary 3rd party actions to do relatively trivial things like downloading protoc or creating a GH release. Hopefully this will help surface errors

This is all that's needed on a non-matrix build (https://sureshjoshi.com/development/github-actions-immutable-release-provenance#release-workflow):

```bash
    - name: Generate
      env:
        GH_TOKEN: ${{ github.token }}
      run: |
        gh release create \
          ${{ needs.determine-version.outputs.release-version }} \
          ${{ needs.run-ci.outputs.artifact-path }} \
          --fail-on-no-commits \
          --generate-notes \
          --repo sureshjoshi/scratch
    
    - name: Generate ${{ needs.determine-version.outputs.release-version }} build provenance attestations
      uses: actions/attest-build-provenance@v3
      with:
        subject-path: ${{ needs.run-ci.outputs.artifact-path }}
```

For a matrix build, I think I just need to run the matrix, create a step that relies on the output of the matrix, and create the release there...

Note: This PR should be the thing that lets me test the CI changes... Just putting reviewers on in case anyone is more familiar with the quirks of action runs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@
 version: 2
 updates:
   - package-ecosystem: "cargo"
+    cooldown:
+      default-days: 7
     directory: "/"
     groups:
       rust-deps:
@@ -13,6 +15,8 @@ updates:
       interval: "weekly"
 
   - package-ecosystem: "github-actions"
+    cooldown:
+      default-days: 7
     directory: "/"
     groups:
       gha-deps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,17 @@ jobs:
     steps:
       - name: Noop
         run: "true"
+
+      # TODO: This is just a temporary thing for debugging
+      - name: Fetch rate_limit JSON
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          curl -s -H "Authorization: token $GITHUB_TOKEN" \
+                   -H "Accept: application/vnd.github+json" \
+                   https://api.github.com/rate_limit \
+            | jq .
+
   ci:
     name: (${{ matrix.name }}) CI
     needs: org-check
@@ -34,8 +45,8 @@ jobs:
           - os: ubuntu-22.04
             name: ubuntu-22.04
 
-          - os: [runs-on, runner=4cpu-linux-arm64, image=ubuntu22-full-arm64-python3.7-3.13, "run-id=${{ github.run_id }}"]
-            name: linux-arm64
+          - os: ubuntu-22.04-arm
+            name: ubuntu-22.04-arm
 
           - os: macos-14
             name: macos-14-arm64
@@ -49,14 +60,18 @@ jobs:
       SCIENCE_AUTH_API_GITHUB_COM_BEARER: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v6
+
       - name: Check Formatting
         run: |
           rustup toolchain add nightly -c rustfmt
           cargo +nightly fmt --check --all
+
       - name: Lint
         run: cargo clippy --locked --all
+
       - name: Unit Tests
         run: cargo test --all
+
       - name: Setup Python 3.9, 3.10, 3.11 (Ubuntu only)
         if: ${{ matrix.os == 'ubuntu-22.04' }}
         uses: actions/setup-python@v6
@@ -65,12 +80,14 @@ jobs:
             3.9
             3.10
             3.11
+
       - name: Compute cache key
         id: build_it_cache_key
         run: |
           # The caches include venvs which have absolute links to Python binaries, so our system
           # should be resilient to this (see `test_pants_source_mode` in `test.rs`).
           echo "cache_key=${{ matrix.os }}-scie-pants-v7-$(which python)" | tee -a "$GITHUB_OUTPUT"
+
       - name: Cache Build and IT Artifacts
         uses: actions/cache@v5
         with:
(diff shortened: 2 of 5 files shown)


sureshjoshi:
Latest update works on my fork with some mods:
https://github.com/sureshjoshi/scie-pants/releases/tag/3.0.0

This doesn't cover announcing to Slack yet, but the process has been simplified overall with Github runners and workflow_calls - so that `ci` builds everything, while `release` is now just responsible for calling that, and then releasing to Github and announcing.

.github/workflows/ci.yml
```diff
@@ -80,47 +95,38 @@ jobs:
...
-              '
-          echo
-          echo "Running under: $(uname -a)"
-          echo
           cargo run -p package -- test \
-            --tools-pex dist/tools.pex --scie-pants dist/scie-pants \
             --check \
             --tools-pex-mismatch-warn
+
+      - name: Package scie-pants ${{ matrix.name }} binary
```
sureshjoshi, line 113: This ensures we can build scie-pants itself, which I think is a good CI check and, if I disable the workflow_dispatch checker in the next step - we could also have "snapshot" full-builds too for a few days as archives

.github/workflows/release.yml
```diff
@@ -53,145 +61,134 @@ jobs:
...
+          path: dist/
 
-      - name: Publish ${{ needs.determine-tag.outputs.release-tag }} Release
-        # See above for discussion:
-        uses: huonw/action-gh-release@998f80d5380609557d7464b01d59a10d845600a0 # v2.0.9 (v2) + https://github.com/softprops/action-gh-release/pull/316
+      - name: Create release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release create \
```
sureshjoshi, line 102: I tried to keep this as close as possible to the kinda convoluted steps we had about publishing earlier from a matrix - now, since we just call out to `ci` and then upload, we're decoupled from the build process.

The `generate-notes` is a copy/paste from my internal repos that I need to probably change up with the changelog.

.github/workflows/release.yml
```diff
@@ -53,145 +61,134 @@ jobs:
...
-              repo: 'homebrew-tap',
-              workflow_id: 'release.yml',
-              ref: 'main',
-              inputs: {
-                tag: '${{ needs.determine-tag.outputs.release-tag }}'
-              }
-            })
+          subject-path: dist/scie-pants-*
+
+  # TODO
```
sureshjoshi: Coming back to this - it's not something I can really test on my fork, and I can only survive so much GHA a day.

sureshjoshi:
Anyone have objections to getting rid of needing to have a release tag before we kick off a release build? Instead, just letting CI build latest off `main`?

It's pretty infrequent.

benjyw:
> Anyone have objections to getting rid of needing to have a release tag before we kick off a release build? Instead, just letting CI build latest off `main`?
> 
> It's pretty infrequent.

Would the release tag come later, or are you proposing to get rid of them entirely?

sureshjoshi:
> > Anyone have objections to getting rid of needing to have a release tag before we kick off a release build? Instead, just letting CI build latest off `main`?
> > It's pretty infrequent.
> 
> Would the release tag come later, or are you proposing to get rid of them entirely?

Nahh, they're generated during the release process. e.g. Created by the act of releasing, not in advance of.

benjyw:
> > > Anyone have objections to getting rid of needing to have a release tag before we kick off a release build? Instead, just letting CI build latest off `main`?
> > > It's pretty infrequent.
> > 
> > 
> > Would the release tag come later, or are you proposing to get rid of them entirely?
> 
> Nahh, they're generated during the release process. e.g. Created by the act of releasing, not in advance of.

Ah, totally fine by me!

.github/workflows/release.yml
```diff
@@ -1,159 +1,140 @@
 name: Release
 on:
-  push:
```
sureshjoshi, line 3: Removed this entirely - the release process is driven via GHA actions (either using Web UI or CLI)

.github/workflows/release.yml
```diff
@@ -1,159 +1,140 @@
...
+
     steps:
       - name: Determine Tag
         id: determine-tag
         run: |
-          if [[ -n "${{ github.event.inputs.tag }}" ]]; then
-            RELEASE_TAG=${{ github.event.inputs.tag }}
-            PRERELEASE=${{ github.event.inputs.prerelease }}
-          else
-            RELEASE_TAG=${GITHUB_REF#refs/tags/}
```
sureshjoshi, line 45: Not needed, as we only drive via GHA workflows now

.github/workflows/release.yml
```diff
@@ -1,159 +1,140 @@
...
-            PRERELEASE=false
-          fi
-          if [[ "${RELEASE_TAG}" =~ ^v[0-9]+.[0-9]+.[0-9]+$ ]] || [[ "${PRERELEASE}" == true ]]; then
+          if [[ "${RELEASE_TAG}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]] || [[ "${PRERELEASE}" == true ]]; then
```
sureshjoshi, line 48: Subtle bugfix here, with escaping the periods - so we can capture them explicitly, rather than any character placeholders.

.github/workflows/release.yml
```diff
@@ -1,159 +1,140 @@
...
-      - name: Package scie-pants ${{ needs.determine-tag.outputs.release-tag }} binary
-        if: ${{ matrix.os == 'ubuntu-22.04' || matrix.name == 'linux-arm64' }}
+          sparse-checkout: |
+            .
+
+      - name: Download artifacts
+        uses: actions/download-artifact@v8
+        with:
+          merge-multiple: true
+          path: dist/
```
sureshjoshi, line 85: Technically this should use the output of `run-ci` but that might be a follow-up PR, need to get a successful scie-pants out in the world with 3.14.

.github/workflows/release.yml
```diff
@@ -1,159 +1,140 @@
...
-      # Build up a draft release with the artifacts from each of these jobs:
-      - name: Create ${{ needs.determine-tag.outputs.release-tag }} Release
-        # Need to pull in https://github.com/softprops/action-gh-release/pull/316 to work-around
-        # double-release-creation that happens when attempting to update a draft release to
-        # not-draft.
-        uses: huonw/action-gh-release@998f80d5380609557d7464b01d59a10d845600a0 # v2.0.9 (v2) + https://github.com/softprops/action-gh-release/pull/316
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Using Markdown headers, grab everything between the current version and subsequent (including current version title)
+          CHANGELOG_NOTES=$(sed -n "/^## ${RELEASE_VERSION}/, /^## /p" CHANGES.md | sed '$d')
```
sureshjoshi, line 95: Replaced the Changelog action with this "probably good enough" one-liner that I use in my repos.

.github/workflows/release.yml
```diff
@@ -1,159 +1,140 @@
...
+    needs: 
+      - determine-tag
+      - run-ci
+    runs-on: ubuntu-latest
     environment: Release
+    permissions:
+      id-token: write
+      attestations: write
+      contents: write
+      discussions: write
```
sureshjoshi, line 72: Need discussion permissions to allow writing to "Announcements"

sureshjoshi:
Alrighty reviewers, this is good to go. I took this a bit further than I had expected - but it appears to now be reliable. 

I have NOT tested the Slack/HomeBrew parts - but we can do those manually as well if there are bugs to fix, the "create new release" part is what matters right now.

Feel free to review my suffering in my fork: https://github.com/sureshjoshi/scie-pants/actions

The output: 
https://github.com/sureshjoshi/scie-pants/releases/tag/0.13.1
https://github.com/sureshjoshi/scie-pants/discussions/7

.github/dependabot.yml
```diff
@@ -4,6 +4,8 @@
 version: 2
 updates:
   - package-ecosystem: "cargo"
+    cooldown:
```
sureshjoshi, line 7: Unrelated to this PR, just a good time to add this (similar to pants itself)

benjyw (approved):
LGTM! Thanks for grinding on this.

cburroughs:
I'm not sure I understand the full consequences of the switch away from `rust:1.94.0-alpine3.23`.  Does that mean instead of static linking of musl, we have a dynamic linking dependency on specific glibc versions?

